PR #7, opened 2018-05-02
Implement conditional logic for role creation in iam-assumable-roles

jkinred:
This allows the advertised variables to function.

The outputs had to be modified to work with conditionals in line with
the workaround in https://github.com/hashicorp/terraform/issues/17425.


antonbabenko:
Good catch, good avatar, merging :)

PR #19, opened 2019-03-15
assumable roles for Users with SAML Identity Provider 

jczerniak:
Module for creating roles for users with SAML Identity Provider like, azure, okta, itp.

### DIFF
--- a/examples/iam-assumable-roles-with-saml/main.tf
+++ b/examples/iam-assumable-roles-with-saml/main.tf
@@ -1,0 +1,26 @@
+provider "aws" {
+  region = "eu-west-1"
+}
+
+resource "aws_iam_saml_provider" "idp_saml" {
+  name                   = "idp_saml"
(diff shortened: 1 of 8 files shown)


examples/iam-assumable-roles-with-saml/main.tf
```diff
@@ -0,0 +1,26 @@
+provider "aws" {
+  region = "eu-west-1"
+}
+
+resource "aws_iam_saml_provider" "idp_saml" {
+  name                   = "idp_saml"
+  saml_metadata_document = "${file("${path.module}/saml-metadata.xml")}"
```
antonbabenko: Could you please add this file also? We need to make sure that examples are always working with just `terraform apply`.

antonbabenko: Fixed, tested.

PR #23, opened 2019-05-09
Additional Policies in Role

ppieprzycki:
I'd like to propose a new feature that allows attaching multiple policies to a role. 
We can now define roles that use multiple policies, it's useful when a user uses AWS defined policies.

### DIFF
--- a/modules/iam-assumable-roles-with-saml/main.tf
+++ b/modules/iam-assumable-roles-with-saml/main.tf
@@ -1,9 +1,3 @@
-locals {
-  admin_additional_policies     = "${length(var.admin_role_additional_policies_arn)}"
-  poweruser_additional_policies = "${length(var.poweruser_role_additional_policies_arn)}"
-  readonly_additional_policies  = "${length(var.readonly_role_additional_policies_arn)}"
-}
-
 data "aws_iam_policy_document" "assume_role_with_saml" {
   statement {
     effect = "Allow"
@@ -44,10 +38,10 @@ resource "aws_iam_role_policy_attachment" "admin" {
 }
 
 resource "aws_iam_role_policy_attachment" "admin_additional_policies" {
-  count = "${var.create_admin_role && local.admin_additional_policies > 0 ? 1 * local.admin_additional_policies  : 0}"
+  count = "${var.create_admin_role ? length(var.admin_role_additional_policies_arns) : 0}"
 
   role       = "${aws_iam_role.admin.name}"
-  policy_arn = "${element(var.admin_role_additional_policies_arn, count.index)}"
+  policy_arn = "${element(var.admin_role_additional_policies_arns, count.index)}"
 }
 
 # Poweruser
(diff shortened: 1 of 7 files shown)


modules/iam-assumable-roles-with-saml/main.tf
```diff
@@ -45,6 +58,13 @@ resource "aws_iam_role_policy_attachment" "poweruser" {
   policy_arn = "${var.poweruser_role_policy_arn}"
 }
 
+resource "aws_iam_role_policy_attachment" "poweruser_additional_policies" {
+  count = "${var.create_poweruser_role && local.poweruser_additional_policies > 0 ? 1 * local.poweruser_additional_policies  : 0}"
```
antonbabenko: Also, simpler, and there is no need in using `local`:
```
count = "${var.create_admin_role ? length(var.poweruser_role_additional_policies_arn) : 0}"
```

modules/iam-assumable-roles-with-saml/variables.tf
```diff
@@ -1,9 +1,11 @@
 variable "provider_name" {
   description = "Name of the SAML Provider"
+  type        = "string"
```
antonbabenko: There is no need in `type = "string"`. Once this module is converted to support Terraform 0.12 we may review this and add types, if necessary.

ppieprzycki:
Hi
I've made changes as advised. Hope it's better 
Regards

modules/iam-assumable-roles-with-saml/main.tf
```diff
@@ -59,10 +53,10 @@ resource "aws_iam_role_policy_attachment" "poweruser" {
 }
 
 resource "aws_iam_role_policy_attachment" "poweruser_additional_policies" {
-  count = "${var.create_poweruser_role && local.poweruser_additional_policies > 0 ? 1 * local.poweruser_additional_policies  : 0}"
+  count = "${var.create_poweruser_role ? length(var.admin_role_additional_policies_arns) : 0}"
```
antonbabenko: Should be `var.poweruser_role_additional_policies_arns`

ppieprzycki:
Ok just to be sure that I correctly understand your intention. 
Should code look similar to this ? 

```
variable "admin_role_policy_arn" {
  default = "arn:123:arn"
}

variable "admin_role_policy_arns" {
  default = []
}


locals {
  admin_role_policy_arns = "${concat(list(var.admin_role_policy_arn),var.admin_role_policy_arns)}"
}


resource "null_resource" "example1" {
  count = "${var.create_admin_role  ? length(local.admin_role_policy_arns) : 0}"
  provisioner "local-exec" {
    command = "echo aaaa"
    #interpreter = ["bash"]
  }
}
```

antonbabenko:
I've just released [v0.5.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v0.5.0) where this feature has been implemented.

@ppieprzycki Thank you for your contribution.

PR #27, opened 2019-05-29
Upgrade to 0.12 syntax

fearoffish:
Although I only ran the upgrade tool, and I only use iam-user at the moment, the upgrade tool seems to have got it right reading through it.

### DIFF
--- a/modules/iam-account/main.tf
+++ b/modules/iam-account/main.tf
@@ -1,21 +1,22 @@
 data "aws_caller_identity" "this" {
-  count = "${var.get_caller_identity}"
+  count = var.get_caller_identity
 }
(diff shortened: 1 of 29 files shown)


antonbabenko:
Lucky users who are using Terraform 0.12 already can use [v2.0.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.0.0) of this module which has been just released!

The rest of the users can use [v1.0.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v1.0.0) which is compatible with Terraform 0.11.

Thanks @fearoffish for the initial pull-request!

PR #29, opened 2019-06-11
Upgraded module to support Terraform 0.12

antonbabenko:
Based on #27.

Closes #27.


antonbabenko:
Lucky users who are using Terraform 0.12 already can use [v2.0.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.0.0) of this module which has been just released!

The rest of the users can use [v1.0.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v1.0.0) which is compatible with Terraform 0.11.

PR #33, opened 2019-08-01
Added description for roles when using iam-group-with-policies

mogaal:
# Description

I got into a limitation regarding set descriptions for roles used inside `iam-group-with-policies` module. 

I added the possibility of set descriptions for roles using a terraform `lookup` function, which doesn't affect the behaviour because it sets to `null` description (basically not set) in case user doesn't specify description field inside the lits(maps)


antonbabenko:
Thanks @mogaal for this PR.

[v2.3.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.3.0) has been just released.

PR #35, opened 2019-08-20
Fix link to iam-assumable-role example in README

miguelaferreira:
# Description

iam-assumable-role example link was pointing to iam-assumable-roles example. This PR fixes that link.


antonbabenko:
Thank you!

Issue #40, opened 2019-10-29
Add description to role


wilkej:
Hey, 
I'd like to use the module but I'm missing the possibility to add a description to a role.

As I'm new to GitHub and I couldn't find information how to contribute. Here is the source code I wanted to add. Could you update the module?

```
variable "role_description" {
  description = "IAM Role description"
  type        = string
  default     = ""
}

resource "aws_iam_role" "this" {
  count = var.create_role ? 1 : 0

  name                 = var.role_name
  path                 = var.role_path
  max_session_duration = var.max_session_duration
  description          = var.role_description
  permissions_boundary = var.role_permissions_boundary_arn

  assume_role_policy = var.role_requires_mfa ? data.aws_iam_policy_document.assume_role_with_mfa.json : data.aws_iam_policy_document.assume_role.json

  tags = var.tags
}

antonbabenko:
Fixed by #45

PR #45, opened 2020-01-09
iam-assumable-role: add description support

LAKostis:
# Description
This PR adds support for custom description in iam-assumable-role. It can be useful in cases where services depend on role description (like keymaker for example).

This is based on https://github.com/terraform-aws-modules/terraform-aws-iam/issues/40


antonbabenko:
Thanks @LAKostis !

[v2.4.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.4.0) of this module has been just released.

Fixed #40.

PR #46, opened 2020-01-15
add instance profile to iam-assumable-role

alemairebe:
# Description

I added the possibility to create an IAM instance profile to "iam-assumable-role" As it only takes 1 extra parameter, and is whitin the scope of this module (IAM)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 <a name="unreleased"></a>
 ## [Unreleased]
 
-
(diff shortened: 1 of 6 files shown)


dmattia:
I was just about to make this PR, glad I checked first! Thanks for doing this @alemairebe

antonbabenko:
Thanks @alemairebe for this addition!

[v2.7.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.7.0) has been just released.

PR #48, opened 2020-01-27
Rename module from "*-iodc" to "*-oidc"

max-rocket-internet:
This was a typo: https://en.wikipedia.org/wiki/OpenID_Connect


antonbabenko:
Great catch, @max-rocket-internet !

[v2.6.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.6.0) has been just released.

PR #50, opened 2020-02-10
feat: Strip https:// from OIDC provider URL if present

chancez:
The OIDC URL returned from the AWS API and https://github.com/terraform-aws-modules/terraform-aws-eks
contains the https:// scheme in the URL, this handles removing it
automatically, as required for correctly creating IAM policies.

### DIFF
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -1,5 +1,6 @@
 locals {
   aws_account_id = var.aws_account_id != "" ? var.aws_account_id : data.aws_caller_identity.current.account_id
+  provider_url   = replace(var.provider_url, "https://", "")
 }
 
 data "aws_caller_identity" "current" {}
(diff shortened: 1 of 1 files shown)


tkent:
I'd also like to vote for this change. The variable name `provider_url` indicates it's a URL (meaning the protocol should be included). However, the module actually needs the provider fully qualified domain name. If the variable name is not changed to `provider_fqdn` (and nobody has asked for this) then this change will avoid confusion.

max-rocket-internet:
@chancez can you rebase

@antonbabenko can you review

chancez:
Rebased

modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -1,5 +1,6 @@
 locals {
   aws_account_id = var.aws_account_id != "" ? var.aws_account_id : data.aws_caller_identity.current.account_id
+  provider_url   = "${replace(var.cluster_oidc_url, "https://", "")}"
```
antonbabenko, line 3: 1. `var.cluster_oidc_url` - where this is defined?

2. Use Terraform 0.12 short syntax by removing `"${...}"` wrapper.

chancez: 1) No idea, must have been from copying this from our own module. PR is so old I can't remember.
2) Fixed.

antonbabenko:
Thanks, @chancez !

[v2.17.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.17.0) has been just released.

PR #78, opened 2020-06-22
Feature: custom inline policy

janschumann:
## Description
Add support for inline policies in addition to custom policy arns.

## Motivation and Context
We would like to migrate from our self maintained IAM modules to this one. The only missing feature is to support custom inline policy definitions. We don't want to maintain IAM policies, as long as they are not shared with multiple roles. Instead we would like to be able to add inline policies to a role with this module.

## How Has This Been Tested?
This is in a state of a proposal. If you would like to add this feature, we would start to use this implementation in our workspaces and then describe the context on how this has been tested.


hobbsh:
I feel like this is a critical feature. Do you need any help getting this through?

PR #82, opened 2020-07-23
fix(Federated-OIDC-ARN): work with regions such as Govcloud

jae-63:
## Description
Override hard-coding of aws partition and use computed partition instead

## Motivation and Context
For example this is an invalid ARN in govcloud:
   arn:aws:iam::431345505690:oidc-provider/oidc.eks.us-gov-east-1.amazonaws.com/id/xyz
This is valid:
  arn:aws-us-gov:iam::431345505690:oidc-provider/oidc.eks.us-gov-east-1.amazonaws.com/id/xyz

Note that this computed partition is used elsewhere within this module, but has been overlooked here.

## Breaking Changes
no

## How Has This Been Tested?
have tested the equivalent logic locally by using terraform's replace() function around the results of this module

### DIFF
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -16,7 +16,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
       type = "Federated"
 
       identifiers = [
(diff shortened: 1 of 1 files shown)


jae-63:
Withdrawing as dup of #83

PR #83, opened 2020-07-24
fix: Allow modules/iam-assumable-role-with-oidc to work in govcloud

bsmedberg-xometry:
by using the current partition instead of hardcoding "aws"

## Description
Currently this module does not work in govcloud because it hardcodes "arn:aws:...", while govcloud uses ARNs of the form "arn:aws-gov". This changes to use the documented technique for getting the current partition.

## Breaking Changes
None.

## How Has This Been Tested?

I have verified in a local Xometry environment that this produces correct and working IAM role definitions for both govcloud and non-govcloud environments.

### DIFF
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -4,6 +4,8 @@ locals {
 
 data "aws_caller_identity" "current" {}
 
+data "aws_partition" "current" {}
+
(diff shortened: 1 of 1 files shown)


antonbabenko:
Thanks, @bsmedberg-xometry !

[v2.16.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.16.0) has been just released.

PR #108, opened 2020-11-09
Update variables.tf

changhyuni:
## Description
Minimum password length.
You can specify the minimum number of characters allowed in an IAM user password. You can type any number from 6 to 128.

## Motivation and Context
This is supported by official document.


antonbabenko:
Hi @changhyuni !

This default value should be customized by the users of this module as necessary individually.

Issue #109, opened 2020-11-10
iam-assumable-roles-with-saml: Allow for multiple provider ids


ravanapel:
### Issue
The aws_iam_policy_document has support for specifying multiple principle federated identifiers like this:
```
"Principal": {
  "Federated": [
    "arn:aws:iam::012345678912:saml-provider/first",
    "arn:aws:iam::012345678912:saml-provider/second"
  ]
},
```

Currently, this module only supports specifying one provider_id:
```
principals {
  type        = "Federated"
  identifiers = [var.provider_id]
}
```
### Request
Would it be possible to support multiple providers by changing the input variables from `string` to list(string)`? I am willing to have a shot at this, but I don't feel confident in estimating the amount of effort it would take.

antonbabenko:
Good idea. You need to change the input variable type and make it backward compatible. See [this module](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-assumable-role-with-oidc/main.tf#L5) for similar evolution.

ravanapel:
Something like this? Will test it at a later moment.

ravanapel:
Thanks!

PR #110, opened 2020-11-10
iam-assumable-roles-with-saml: Allow for multiple provider ids

ravanapel:
## Description
The aws_iam_policy_document has support for specifying multiple principle federated identifiers like this:
```
"Principal": {
  "Federated": [
    "arn:aws:iam::012345678912:saml-provider/first",
    "arn:aws:iam::012345678912:saml-provider/second"
  ]
},
```

Currently, this module only supports specifying one provider_id:
```
principals {
  type        = "Federated"
  identifiers = [var.provider_id]
}
```

## Motivation and Context
Solves issue #109 

## Breaking Changes
-

## How Has This Been Tested?
Not yet tested

### DIFF
--- a/modules/iam-assumable-roles-with-saml/main.tf
+++ b/modules/iam-assumable-roles-with-saml/main.tf
@@ -1,3 +1,16 @@
+locals {
+  aws_account_id = var.aws_account_id != "" ? var.aws_account_id : data.aws_caller_identity.current.account_id
+  ids            = compact(distinct(concat(var.provider_ids, [var.provider_id])))
+  identifiers = [
+    for id in local.ids :
+    "arn:${data.aws_partition.current.partition}:iam::${local.aws_account_id}:saml-provider/${id}"
+  ]
+}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_partition" "current" {}
+
 data "aws_iam_policy_document" "assume_role_with_saml" {
   statement {
     effect = "Allow"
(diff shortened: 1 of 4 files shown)


modules/iam-assumable-roles-with-saml/main.tf
```diff
@@ -1,12 +1,26 @@
+locals {
+  aws_account_id = var.aws_account_id != "" ? var.aws_account_id : data.aws_caller_identity.current.account_id
+  ids            = compact(distinct(concat(var.provider_ids, [var.provider_id])))
+  identifiers = [
+    for id in local.ids :
```
antonbabenko, line 5: I don't think you need to replace partition and account id in a loop like this. The values in `provider_ids` and/or `provider_id` will be already correct.

ravanapel: You're right. I assumed it was needed because of the other implementation. I've removed it.

antonbabenko:
Could you please also update the code in `examples` for this module?

ravanapel:
Sure! Do you mean expanding the current example with a module that uses multiple provider ids?

antonbabenko:
Yes, update the example to show how one and multiple provider ids can be used.

ravanapel:
I've updated the example.

antonbabenko:
[v3.4.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v3.4.0) has been just released.

Issue #112, opened 2020-11-14
purpose of number_of_role_policy_arns and number_of_custom_role_policy_arns?


Justin-DynamicD:
Question in the title, but I don't see a usable purpose for these two settings?  WHat's the usecase?

Currently the value is just for a count, but as Terraform is declarative, providing the count won't guarantess _which_ role policies get applied, just that "n" will.  So unless you match this value to the coresponding list values, you end up wit han effectively random deployment.

With that in mind, it seems more reliable to simply supply this with length(var.role_policy_arns), and ommit these entires entirely, unless there's a purpose Im not seeing?

antonbabenko:
Look at the discussion in #95 for the explanation.

In short, Terraform doesn't support computed values in the count so we need to tell precisely the number of resources to create.

Justin-DynamicD:
Is this a limitation in older Terraform?  Because in 0.13 that statement is simply untrue:

```
locals {
  sample = ["thing1", "thing2"]
}

resource "local_file" "catinthehat" {
    count = length(local.sample)
    content     = local.sample[count.index]
    filename = "${path.module}/foo-${count.index}.bar"
}
```

and the output:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # local_file.mystery[0] will be created
  + resource "local_file" "catinthehat" {
      + content              = "thing1"
      + directory_permission = "0777"
      + file_permission      = "0777"
      + filename             = "./foo-0.bar"
      + id                   = (known after apply)
    }

  # local_file.mystery[1] will be created
  + resource "local_file" "catinthehat" {
      + content              = "thing2"
      + directory_permission = "0777"
      + file_permission      = "0777"
      + filename             = "./foo-1.bar"
      + id                   = (known after apply)
    }

Plan: 2 to add, 0 to change, 0 to destroy
```

Honestly this isn't really the best way to do it, I would rather do it with a for_each which will allow updates to be a bit more specific so people can add/remove from the list and only create changes for those elements that change:

```
locals {
  sample = ["thing1", "thing2"]
}

resource "local_file" "catinthehat" {
    for_each = toset(local.sample)
    content  = each.value
    filename = "${path.module}/file-${each.key}.bar"
}
```

Issue #95, opened 2020-09-24
terraform iam_assumable_role adaption


Zyntogz:
Hey,

we found an "issue" that you already found and implemented in your lambda module: When using the variable "custom_role_policy_arns" the count parameter will be constructed depending on the length of the list contained in "custom_role_policy_arns". This results in an error of terraform that the length of resources cannot be constructed during runtime. For solution of this, i propose to add a new variable and use this during the count statement:

variables.tf
variable "number_custom_role_policy_arns" {
  description = "Number of IAM policies to attach to IAM role"
  type        = number
  default     = 0
}

main.tf
resource "aws_iam_role_policy_attachment" "custom" {
  count = var.create_role ? var.number_custom_role_policy_arns : 0

  role       = aws_iam_role.this[0].name
  policy_arn = element(var.custom_role_policy_arns, count.index)
}

Thanks!

antonbabenko:
Hi again :)

I see what you mean and it makes sense (unfortunately, Terraform is not so smart yet). Pull-request is welcome.

Zyntogz:
Hey :)

i would love to do so, however i cannot push to a new origin branch. What i do:

git push -u origin feature/iam_assumable_role_number_of_policies_fix
...
remote: Permission to terraform-aws-modules/terraform-aws-iam.git denied to Zyntogz.
fatal: unable to access 'https://github.com/terraform-aws-modules/terraform-aws-iam.git/': The requested URL returned error: 403
...

I can do whatever i want, if https or ssh, it stays like that... You may have any idea?

antonbabenko:
You need to fork the repo, make changes, and send a pull request to this repository.

antonbabenko:
Fixed in #96

[v2.26.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v2.26.0) has been just released.

yogeek:
Hello @antonbabenko 

This fix provoked the deletion of hundreds of role_policy_attachments when our terraform automatically retrieved the last version of this module... indeed, as the variable value is "0" by default, our terraform took an empty array and delete all the attachments.
It implies a pretty big number of problems in our multiple account ^^

In our case, we have an fixed ARNs list and we did not have the error mentionned above.
Do you think this default value to 0 can be adapted to avoid such inconvenience ?

antonbabenko:
[...]

PR #116, opened 2020-11-22
Fix policy counter

Justin-DynamicD:
## Description
changes the `number_of_policy_arns` from a var to a local to remove the need to set this value manually

This is for your consideration, feel free to toss.

## Motivation and Context
having to declare a count of policies is error prone for the end user as well as a breaking change in version from previous versions. By moving this value into a local, we can calculate the number with length() before the resource is created avoiding the potential race condition that caused the previous error to occur while maintaining legacy compatibility.

## Breaking Changes
This reverts the latest change by allowing it to behave predicatbly and thus the extra variables unnessisary. We could leave the old var in place as a "do nothing", but I figured that would confuse things worse.

## How Has This Been Tested?
This has been tested in a local fork, albiet at a small scale (only 6 resources across 3 clusters use this module currently).

### DIFF
--- a/modules/iam-assumable-role-with-oidc/locals.tf
+++ b/modules/iam-assumable-role-with-oidc/locals.tf
@@ -1,0 +1,3 @@
+locals {
+  number_of_role_policy_arns = length(var.role_policy_arns)
+}
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -68,7 +68,7 @@ resource "aws_iam_role" "this" {
 }
 
 resource "aws_iam_role_policy_attachment" "custom" {
-  count = var.create_role ? length(var.role_policy_arns) : 0
+  count = var.create_role ? local.number_of_role_policy_arns : 0
 
   role       = join("", aws_iam_role.this.*.name)
   policy_arn = var.role_policy_arns[count.index]
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -68,7 +68,7 @@ resource "aws_iam_role" "this" {
 }
 
 resource "aws_iam_role_policy_attachment" "custom" {
-  count = var.create_role ? var.number_of_role_policy_arns : 0
+  count = var.create_role ? length(var.role_policy_arns) : 0
 
   role       = join("", aws_iam_role.this.*.name)
   policy_arn = var.role_policy_arns[count.index]
--- a/modules/iam-assumable-role-with-oidc/variables.tf
+++ b/modules/iam-assumable-role-with-oidc/variables.tf
@@ -70,13 +70,6 @@ variable "role_policy_arns" {
   default     = []
 }
 
-variable "number_of_role_policy_arns" {
-  description = "Number of IAM policies to attach to IAM role"
-  type        = number
-  default     = 0
-}
-
-
 variable "oidc_fully_qualified_subjects" {
   description = "The fully qualified OIDC subjects to be added to the role policy"
   type        = set(string)
--- a/modules/iam-assumable-role/locals.tf
+++ b/modules/iam-assumable-role/locals.tf
@@ -1,0 +1,3 @@
+locals {
+  number_of_custom_role_policy_arns = length(var.custom_role_policy_arns)
+}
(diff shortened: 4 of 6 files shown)


antonbabenko:
This has been discussed in multiple issues already (#95 , #112).

The code you have in [here](https://github.com/terraform-aws-modules/terraform-aws-iam/issues/112#issuecomment-729440951) works well:

```hcl
locals {
  sample = ["thing1", "thing2"]   # <= static values
}

resource "local_file" "catinthehat" {
    count = length(local.sample)
    content     = local.sample[count.index]
    filename = "${path.module}/foo-${count.index}.bar"
}
```

But this one won't work without firstly targeting resources used in `count` or `for_each` (eg, `terraform apply -target=random_pet.this -target=random_pet.that`):

```hcl
resource "random_pet" "this" {}
resource "random_pet" "that" {}

locals {
  sample = [random_pet.this.id, random_pet.that.id]  # <= computed/dynamic values
}

resource "local_file" "catinthehat" {
    count = length(local.sample)
    content     = local.sample[count.index]
    filename = "${path.module}/foo-${count.index}.bar"
}
```

lorengordon:
@antonbabenko fwiw, that code now works on the first apply... At some point in the tf 0.12.x releases, it learned how to tell in the plan stage that the list has two items in it.

```
$ rm -rf .terraform/ terraform.tfstate*
$ cat main.tf
resource "random_pet" "this" {}
resource "random_pet" "that" {}

locals {
  sample = [random_pet.this.id, random_pet.that.id]  # <= computed/dynamic values
}

resource "local_file" "catinthehat" {
    count = length(local.sample)
    content     = local.sample[count.index]
    filename = "${path.module}/foo-${count.index}.bar"
}

$ terraform init && terraform apply -auto-approve

Initializing the backend...

Initializing provider plugins...
- Finding latest version of hashicorp/random...
- Finding latest version of hashicorp/local...
- Using hashicorp/random v3.0.0 from the shared cache directory
- Using hashicorp/local v2.0.0 from the shared cache directory

The following providers do not have any version constraints in configuration,
so the latest version was installed.

To prevent automatic upgrades to new major versions that may contain breaking
changes, we recommend adding version constraints in a required_providers block
in your configuration, with the constraint strings suggested below.

* hashicorp/local: version = "~> 2.0.0"
* hashicorp/random: version = "~> 3.0.0"

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.
random_pet.this: Creating...
random_pet.that: Creating...
random_pet.this: Creation complete after 0s [id=composed-sawfly]
random_pet.that: Creation complete after 0s [id=assured-gelding]
local_file.catinthehat[1]: Creating...
local_file.catinthehat[0]: Creating...
local_file.catinthehat[1]: Creation complete after 0s [id=7e392649803da5af9a49e9b640ba4e65f4bf286e]
local_file.catinthehat[0]: Creation complete after 0s [id=965ce6f938f93b2753394c2ff144ce37a3889555]

Apply complete! Resources: 4 added, 0 changed, 0 destroyed.
```

lorengordon:
It even works when the length of the list is dynamic, as long as the number of items can be determined entirely from variables or data sources that are resolved in the plan phase.

antonbabenko:
You are right, this is an interesting observation! There are still some situations when the number of resources can't be identified by Terraform and an error described in #95 happens.

The main question I still have not figured out a solution for is how to make it work for both cases? Will this work `try(length(var.something), var.number_of_something)`?

I don't have time to work on this myself, but maybe someone can verify and propose a universal solution

lorengordon:
I don't think it is possible to cover every possible crazy thing a user might do. I mean, even now, they could have a dynamically generated list that terraform can't compute on first apply, and try to pass `number_of_something = length(<dynamic list>)`. At some point, we have to just say sorry terraform doesn't support that workflow.

I think these days terraform is doing an awful lot to minimize the use cases where the count cannot be computed, and the rest is up to the user to understand the one limitation left.

Justin-DynamicD:
Thank you for clarifying the scenario in which it causes failure.  I'll probably do some additional testing to see if I can squash all scenarios, but frankly I cant make it fail on TF 0.13.4 so part of me is starting to wonder if the proper solution may be cranking up the minium version rather than over-engineering ;)

I'll try an earlier version and see if I can do somethign that isnt overly-crusty. Again, feel free to reject if Im going inu wierd directions.

yukin01:
I think #119 would work in both cases. What do you think?

antonbabenko:
[v3.6.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v3.6.0) has been just released.

PR #119, opened 2020-11-26
fix: automatically determine the number of role policy arns

yukin01:
## Description

Changed to automatically calculate `number_of_role_policy_arns` even if you omit this.

Added backward compatibility to #116

## Motivation and Context

By default the `number_of_role_policy_arns` should be calculated automatically, but it can be also specified explicitly as variable.

## Breaking Changes

Probably not

## How Has This Been Tested?

Use new example.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 terraform.tfstate
 *.tfstate*
 terraform.tfvars
+.terraform.lock.hcl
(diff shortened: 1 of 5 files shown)


antonbabenko:
Thanks @yukin01 !

[v3.5.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v3.5.0) has been just released.

PR #122, opened 2020-12-23
feat: add github action workflow to execute pre-commit checks

bryantbiggs:
## Description
- update pre-commit hook versions utilized
- add GitHub action workflows for running static checks using pre-commit

## Motivation and Context
- introducing static check CI/CD workflow to project repositories

## Breaking Changes
- No

## How Has This Been Tested?
- No material changes to source code


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,69 @@
...
+    needs: getBaseVersion
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - ${{ needs.getBaseVersion.outputs.minVersion }}
+          - ${{ needs.getBaseVersion.outputs.maxVersion }}
+        directory:
+          - '.'
+          - modules/iam-account
```
antonbabenko, line 37: It would be great if this list can be autogenerated from the actual code. We already have some conversions, eg code is in the root and in `modules/*`.

I can't imagine anyone updating these workflow files manually correctly.

.github/workflows/examples.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Install Terraform MAX v${{ steps.minMax.outputs.maxVersion }}
+        uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_version: ${{ steps.minMax.outputs.maxVersion }}
+
+      - name: Validate MAX v${{ steps.minMax.outputs.maxVersion }}
+        run: |
+          cd ${{ matrix.directory }}
+          terraform init -backend=false
+          terraform validate
```
antonbabenko, line 58: I think we can just include `examples/*` into `pre-commit`, because `terraform validate` is already part of `pre-commit`.

PR #139, opened 2021-03-04
fix: handle unencrypted secrets

nikolay:
## Description
There are use cases when encryption is not possible and these are not handled properly right now.

## Motivation and Context
Sometimes internal accounts are created and they never leave Terraform and encryption is not possible due to the lack of decryption functionality within Terraform.

## Breaking Changes
When not appropriate, Keybase commands will just be null, which could be extended further to check if a regular PGP key is used and not a Keybase one.


antonbabenko:
Looks good! Thank you for this improvement!

[v3.11.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v3.11.0) has been just released.

PR #148, opened 2021-04-15
fix: Set sensitive=true for sensitive outputs and use tolist()

### LINKED ISSUE #146: Prepare code for 0.15
0.15 was released and in order to use modules here we'd need to remove deprecated syntax like: https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-assumable-role/main.tf#L2

antonbabenko:
Fixes #146


antonbabenko:
[v3.15.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v3.15.0) has been just released.

PR #155, opened 2021-06-02
fix(iam-user): Add 'sensitive' flag to output 'pgp_key'

carnei-ro:
## Description

Add "sensitive" flag to output "pgp_key" for module iam-user.  
Fix `Error: Output refers to sensitive values`.

## Motivation and Context

Fix `Error: Output refers to sensitive values`.

## Breaking Changes

None

## How Has This Been Tested?

### DIFF
--- a/modules/iam-user/outputs.tf
+++ b/modules/iam-user/outputs.tf
@@ -83,6 +83,7 @@ output "iam_access_key_status" {
 
 output "pgp_key" {
   description = "PGP key used to encrypt sensitive data for this user (if empty - secrets are not encrypted)"
+  sensitive   = true
   value       = var.pgp_key
 }
 


modules/iam-user/outputs.tf
```diff
@@ -83,6 +83,7 @@ output "iam_access_key_status" {
 
 output "pgp_key" {
   description = "PGP key used to encrypt sensitive data for this user (if empty - secrets are not encrypted)"
+  sensitive   = true
   value       = var.pgp_key
```
antonbabenko, line 87: `pgp_key` is input into a module and I don't think it makes a lot of sense to mark it sensitive here. It makes sense to mark `sensitive = true` only resource attributes (see line 46 in this file).

carnei-ro: hmmm I agree with you, although I am getting `Error: Output refers to sensitive values.` when I am trying to use it. (Terraform 0.14.3 / Terragrunt v0.28.19)

carnei-ro: Nevermind, it seems to be not related to this module itself, but the fact I am creating a new module and using this one as one of the sources....

PR #163, opened 2021-09-10
feat: Add ability for controlling whether or not to create a policy

gabops:
## Description
Add ability for controlling whether or not to create a policy

## Motivation and Context
In certain scenarios can be useful to decide which specific policy should be created or not. Let's imagine that we create a module from this module for creating a set of policies that might change depending on the account you are targeting. This change allows you to do that.

## Breaking Changes
There is no breaking change as the module still behaving the same by default.

## How Has This Been Tested?
- [✅] I have tested and validated these changes using one or more of the provided `examples/*` projects
- [✅] Tested over my own AWS account.

### DIFF
--- a/examples/iam-policy/main.tf
+++ b/examples/iam-policy/main.tf
@@ -53,3 +53,9 @@ module "iam_policy_from_data_source" {
     PolicyDescription = "Policy created using example from data source"
   }
 }
+
+module "iam_policy_optional" {
+  source = "../../modules/iam-policy"
+
+  create_policy = false
+}
--- a/examples/iam-policy/README.md
+++ b/examples/iam-policy/README.md
@@ -34,6 +34,7 @@ Run `terraform destroy` when you don't need these resources.
 |------|--------|---------|
 | <a name="module_iam_policy"></a> [iam\_policy](#module\_iam\_policy) | ../../modules/iam-policy |  |
 | <a name="module_iam_policy_from_data_source"></a> [iam\_policy\_from\_data\_source](#module\_iam\_policy\_from\_data\_source) | ../../modules/iam-policy |  |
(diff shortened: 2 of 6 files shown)


examples/iam-policy/main.tf
```diff
@@ -53,3 +53,28 @@ module "iam_policy_from_data_source" {
     PolicyDescription = "Policy created using example from data source"
   }
 }
+
+module "iam_policy_optional" {
+  source        = "../../modules/iam-policy"
+  create_policy = false
+
+  name        = "example_optional"
```
antonbabenko, line 61: There are `default` values already specified for all variables in the module, so you can remove everything starting from line 61 ([like here](https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/master/examples/complete-vpc/main.tf#L171-L175)).

antonbabenko (changes requested):

examples/iam-policy/main.tf
gabops: Ok, I've removed "path" and "description".

Thank you for your time!

antonbabenko: Please remove everything and leave just `source` and `create_policy = false` with empty line in between.

gabops: Done.

antonbabenko:
Thanks, @gabops !

[v4.4.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v4.4.0) has been just released.

gabops:
> Thanks, @gabops !
> 
> [v4.4.0](https://github.com/terraform-aws-modules/terraform-aws-iam/releases/v4.4.0) has been just released.

Thanks to you for your time reviewing this!

PR #169, opened 2021-10-01
adding override_json variable to data policy document and adding trusted_role_action to iam-assumable-roles

tcharewicz:
## Description
Adding variable assume_role_policy, and use it at data policy document level as a override_json.

## Motivation and Context
You can create any proper assume policy document and pass it to the IAM module.

## Breaking Changes
None

## How Has This Been Tested?
Right now this changes arn't tested with examples presented at repository. If maintainer accept this way of configuration I will make this tests.

### DIFF
--- a/examples/iam-account/README.md
+++ b/examples/iam-account/README.md
@@ -30,7 +30,7 @@ No providers.
 
 | Name | Source | Version |
 |------|--------|---------|
-| <a name="module_iam_account"></a> [iam\_account](#module\_iam\_account) | ../../modules/iam-account |  |
+| <a name="module_iam_account"></a> [iam\_account](#module\_iam\_account) | ../../modules/iam-account | n/a |
 
 ## Resources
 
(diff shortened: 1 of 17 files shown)


fn-jbarker:
This would have been a nice feature for me recently.  The either or scenario with mfa didn't allow me to entertain the scenario where we have humans interacting with the role in addition to a service account interacting with the role.  Both needed to assume the role, but the service account isn't required to be mfa authenticated.

fn-jbarker:
`{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::000000000000:root"
      },
      "Action": "sts:AssumeRole",
      "Condition": {
        "Bool": {
          "aws:MultiFactorAuthPresent": "true"
        }
      }
    },
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::000000000000:role/madeuprole"
      },
      "Action": "sts:AssumeRole"
    }
  ]
}`

fn-jbarker:
Thoughts on this @antonbabenko?

antonbabenko (changes requested):
Looks very good (skip CI failures, I will fix those after this PR is done). Could you update code in the examples for the affected modules to show `assume_role_policy` in action?

PR #170, opened 2021-10-13
WIP: Adds role_allows_session_tag

### LINKED ISSUE #152: iam-assumable-role: support sts:TagSession or custom assume policy
The default behavior for the GitHub Action "configure-aws-credintials" requires sts: TagSession permissions, without Condition

- Add a variable for custom assume policy json
- Add a flag for add sts:TagSession

Trust Relationship Example
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "",
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::*:user/test_user"
      },
      "Action": "sts:AssumeRole",
      "Condition": {
        "StringEquals": {
          "sts:ExternalId": "TestID"
        }
      }
    },
    {
      "Sid": "",
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::*:user/test_user"
      },
      "Action": "sts:TagSession"
    }
  ]
}
```

## Additional context
https://github.com/aws-actions/configure-aws-credentials#permissions-for-assuming-a-role
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html

RafPe:
## Description
Change allows to control via parameter if sts:TagSession statement is added into role assume document 

## Motivation and Context
The change is required for a lot of people using GithubActions and relying on this repository structures for controlling IAM. Without this additional workarounds are required to fully use the solution. 
Fix #152 

## Breaking Changes
There are no breaking changes

## How Has This Been Tested?

### DIFF
--- a/modules/iam-assumable-role/README.md
+++ b/modules/iam-assumable-role/README.md
@@ -56,6 +56,7 @@ No modules.
 | <a name="input_role_name"></a> [role\_name](#input\_role\_name) | IAM role name | `string` | `""` | no |
 | <a name="input_role_path"></a> [role\_path](#input\_role\_path) | Path of IAM role | `string` | `"/"` | no |
 | <a name="input_role_permissions_boundary_arn"></a> [role\_permissions\_boundary\_arn](#input\_role\_permissions\_boundary\_arn) | Permissions boundary ARN to use for IAM role | `string` | `""` | no |
+| <a name="input_role_allows_session_tag"></a> [role\_allows\_session\_tag](#input\_role\_allows\_session\_tag) | Whether role creates IAM statement to allow sts:TagSession | `bool` | `false` | no |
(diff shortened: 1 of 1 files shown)


RafPe:
> > @lorengordon I think that is exactly what we should be doing! Spot on :) And since we are on the subject should we not also update [`master`/modules/iam-assumable-role/main.tf#L36](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-assumable-role/main.tf#L36) with trusted_role_actions ?
> 
> Makes sense to me, but I'd defer to Anton. I feel like that's probably the only change needed, and the rest can be undone?
Ok - lets wait for input from Anton - then it would be just making a new small patch PR. For the rest I will close the PR and we can comment on the linked issue that the solution is there out of the box

antonbabenko:
I agree, let's make it `trusted_role_actions` as variable with current value as default.

PR #174, opened 2021-11-16
feat: Added iam-read-only-policy module

Andrey9kin:
## Description
Added new module to create customizable read only policy

## Motivation and Context
* As stated in the readme: Creates IAM read-only policy for specified services. Default AWS read-only policies (arn:aws:iam::aws:policy/job-function/ViewOnlyAccess, arn:aws:iam::aws:policy/ReadOnlyAccess), being a one-size-fits-all type of policies, have a lot of things missing as well as something that you might not need. Also, AWS default policies are known for having [security issues](https://securityboulevard.com/2020/12/the-aws-managed-policies-trap/)
Thus this module is an attempt to build a better base for a customizable usable read-only policy.
* Module requires a list of allowed services which forces you to compile such a list which will be very useful if you are to create an SCP to forbid usage of not allowed services.
* Most of our customers asked for something like that so their developers can login to prod with proper read only access and we had a number of requests to have such module available as open source
* We could open source it under fivexl org but here feels like a better place. Hopefully more people will be able to take advantage of it and perhaps contribute back. So win-win all around

## Breaking Changes
A new thing. Does not break anything

## How Has This Been Tested?
- Battle-tested in a several projects


Andrey9kin:
@antonbabenko could you please take a look when you have time or ask someone to do so. Thanks!

examples/iam-read-only-policy/main.tf
```diff
@@ -0,0 +1,39 @@
...
+    local.allowed_services_storage,
+    local.allowed_services_databases,
+    local.allowed_services_management,
+    local.allowed_services_analytics,
+    local.allowed_services_application,
+    local.allowed_services_security
+  )
+}
+
+module "iam_policy" {
```
antonbabenko, line 27: Add another example of module call with `create_policy = false` and update variables defaults accordingly.

examples/iam-read-only-policy/main.tf
```diff
@@ -0,0 +1,39 @@
...
+  allowed_services_compute     = ["ec2", "ecr", "eks", "ecs", "lambda", "autoscaling", "elasticloadbalancing"]
+  allowed_services_networking  = ["vpc", "route53", "route53domains", "route53resolver", "servicediscovery"]
+  allowed_services_storage     = ["s3", "backup", "dynamo", "dms", "elasticfilesystem"]
+  allowed_services_databases   = ["rds", "dynamo", "elasticache"]
+  allowed_services_management  = ["cloudwatch", "events", "logs", "servicequotas", "ssm"]
+  allowed_services_analytics   = ["es", "firehose", "kinesis", "kinesisanalytics", "redshift"]
+  allowed_services_application = ["ses", "sns", "sqs", "xray", "applicationinsights", "application-autoscaling"]
+  allowed_services_security    = ["iam", "acm", "kms", "secretsmanager"]
+
+  allowed_services = concat(
```
antonbabenko, line 15: `distinct(concat(...))` to be safer :)

examples/iam-read-only-policy/versions.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  required_version = ">= 0.15.0"
```
antonbabenko, line 2: Please set `>= 0.12.6` as on all other modules. We will bump to `>= 0.13.1` in all modules in another PR.

examples/iam-read-only-policy/versions.tf
```diff
@@ -0,0 +1,7 @@
...
+
+  required_providers {
+    aws = ">= 2.23"
```
antonbabenko, line 5: If I remember correctly, 2.23 does not work with Terraform 0.15.

Please downgrade the minimum version of Terraform (see comment above).

modules/iam-read-only-policy/README.md
```diff
@@ -0,0 +1,56 @@
+# iam-policy
```
antonbabenko, line 1:
```suggestion
# iam-read-only-policy
```

modules/iam-read-only-policy/main.tf
```diff
@@ -0,0 +1,90 @@
...
+
+  tags = var.tags
+}
+
+locals {
+  console_services = ["resource-groups", "tag", "health"]
+}
+
+data "aws_iam_policy_document" "allowed_services" {
+
```
antonbabenko, line 18: No resources or data sources should be created if `create_policy = false` is set. Add `count` to everything.

modules/iam-read-only-policy/outputs.tf
```diff
@@ -0,0 +1,9 @@
+output "policy_object" {
+  description = "Policy object if created"
+  value       = one(aws_iam_policy.policy[*])
```
antonbabenko, line 3: There is no need to use 0.15 features here.

Also, it is rather a bad idea to expose the whole resource with all the properties (same as in programming in most cases). Please change to export attributes individually.

modules/iam-read-only-policy/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+  type        = string
+  default     = "IAM Policy"
+}
+
+variable "allowed_services" {
+  description = "List of services to allow Get/List/Describe/View options. Service name should be the same as corresponding service IAM prefix. See what it is for each service here https://docs.aws.amazon.com/service-authorization/latest/reference/reference_policies_actions-resources-contextkeys.html"
+  type        = list(string)
+}
+
+variable "additional_policy" {
```
antonbabenko, line 30:
```suggestion
variable "additional_policy_json" {
```

This is how such variable is called in other modules (e.g. Lambda, Step Functions).

modules/iam-read-only-policy/main.tf
```diff
@@ -0,0 +1,90 @@
...
+  path        = var.path
+  description = var.description
+
+  policy = data.aws_iam_policy_document.combined.json
+
+  tags = var.tags
+}
+
+locals {
+  console_services = ["resource-groups", "tag", "health"]
```
antonbabenko, line 14: Let's move these from locals to separate variable defaults to allow customizations (if necessary).

modules/iam-read-only-policy/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+}
+
+variable "allow_predefined_sts_actions" {
+  description = "Allows GetCallerIdentity/GetSessionToken/GetAccessKeyInfo sts actions"
+  type        = bool
+  default     = true
+}
+
+variable "allow_web_console_services" {
+  description = "Allows read actions for resource-groups, tag, health services that needed when browing AWS console"
```
antonbabenko, line 55:
```suggestion
  description = "Allows List/Get/Describe/View actions for services used when browsing AWS console (e.g. resource-groups, tag, health services)"
```

Similar to the descriptions of other vars.

modules/iam-read-only-policy/versions.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  required_version = ">= 0.15.0"
```
antonbabenko, line 2: Same version requirements here as in the example.

examples/iam-read-only-policy/README.md
```diff
@@ -0,0 +1,53 @@
+# IAM user example
```
antonbabenko, line 1:
```suggestion
# IAM read-only policy example
```

antonbabenko (changes requested):
Please add relevant updates in the main README in the "Usage" and "Examples" sections.

RafPe:
@Andrey9kin any update on completing this resource ?  I can see potential on using it.

Andrey9kin:
@RafPe will be looking into it tonight! Got few more people asking for it as well

examples/iam-read-only-policy/main.tf
Andrey9kin: added distinct to the module itself to sanitize input

Andrey9kin: thus here it will be unnecessary

examples/iam-read-only-policy/versions.tf
Andrey9kin: done

examples/iam-read-only-policy/versions.tf
Andrey9kin: done

modules/iam-read-only-policy/main.tf
Andrey9kin: intention of create_policy is to skip creation of iam policy and only generate json doc thus you need this data in both cases and that is why there is no count

modules/iam-read-only-policy/outputs.tf
Andrey9kin: could argue with that in this particular case but alright will do as you suggested

modules/iam-read-only-policy/versions.tf
Andrey9kin: done

modules/iam-read-only-policy/main.tf
Andrey9kin: done

Andrey9kin:
@antonbabenko sorry for keeping it hanging but end of the year was kind of crazy. Thanks for the great review comments. When you are back from holidays please take a look. Thanks!

modules/iam-read-only-policy/main.tf
Andrey9kin: if I misunderstood your comment please give me an example of what you mean

antonbabenko: Normally, in other terraform-aws-modules, nothing is being created if `create_* = false` is defined. This way is compatible with situations when users want to control the creation of `all` or `nothing` (e.g. still handy with terragrunt).

I see what you mean in this example. Let's leave it as you have done already. I think it makes sense here.

modules/iam-read-only-policy/versions.tf
```diff
@@ -0,0 +1,7 @@
+terraform {
+  required_version = ">= 0.12.6"
+
+  required_providers {
+    aws = ">= 3.35"
```
antonbabenko, line 5:
```suggestion
    aws = ">= 2.23"
```

examples/iam-read-only-policy/README.md
```diff
@@ -0,0 +1,59 @@
+# IAM read-only policy example
+
+Configuration in this directory creates read only IAM policy.
```
antonbabenko, line 3:
```suggestion
Configuration in this directory creates read-only IAM policy and attaches it to AWS SSO.
```

antonbabenko (approved):

antonbabenko:
This PR is included in version 4.8.0 :tada:

PR #176, opened 2021-12-16
feat: Allow setting custom trust policy in iam-assumable-role

simonweil:
## Description
This change adds an option to set a custom trust policy.

## Motivation and Context

I need to create an assumable role that can be assumed from all member accounts of the organization except for sandbox accounts.
This can be achieved via conditions and a deny policy for the sandbox.
As there is no possibility to cover all options of the JSON policy, the best would be for such rare situations to allow a custom trust policy.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-assumable-role/main.tf
+++ b/modules/iam-assumable-role/main.tf
@@ -3,6 +3,8 @@ locals {
 }
 
 data "aws_iam_policy_document" "assume_role" {
+  count = var.custom_role_trust_policy == "" && var.role_requires_mfa ? 0 : 1
+
   statement {
     effect = "Allow"
 
@@ -30,6 +32,8 @@ data "aws_iam_policy_document" "assume_role" {
 }
 
 data "aws_iam_policy_document" "assume_role_with_mfa" {
+  count = var.custom_role_trust_policy == "" && var.role_requires_mfa ? 1 : 0
+
   statement {
     effect = "Allow"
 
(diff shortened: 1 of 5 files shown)


modules/iam-assumable-role/main.tf
```diff
@@ -79,7 +79,7 @@ resource "aws_iam_role" "this" {
   force_detach_policies = var.force_detach_policies
   permissions_boundary  = var.role_permissions_boundary_arn
 
-  assume_role_policy = var.role_requires_mfa ? data.aws_iam_policy_document.assume_role_with_mfa.json : data.aws_iam_policy_document.assume_role.json
+  assume_role_policy = var.custom_role_trust_policy != "" ? var.custom_role_trust_policy : var.role_requires_mfa ? data.aws_iam_policy_document.assume_role_with_mfa.json : data.aws_iam_policy_document.assume_role.json
```
antonbabenko, line 82: Something like this is more in spirit of Terraform 0.13+:
```suggestion
  assume_role_policy = coalesce(var.custom_role_trust_policy, try(data.aws_iam_policy_document.assume_role_with_mfa[0].json, data.aws_iam_policy_document.assume_role[0].json))
```

antonbabenko: Also, add `count = var.role_requires_mfa ? 1 : 0` (and opposite) in both data sources.

antonbabenko (changes requested):
Suggested some improvements

modules/iam-assumable-role/main.tf
simonweil: Thanks for the quick review, will try to fix it all today

simonweil: Done both and fixed doc failed check

simonweil:
Please check again

simonweil:
Testing it with all my setups and role combinations all work fine with no changes detected in plan so I hope we are good :)

simonweil:
Pushed a fix for the failing check, should be all green now.
Tell me if there is anything else to fix

Once you approve, I'll squash fixup commits

antonbabenko:
This PR is included in version 4.10.0 :tada:

simonweil:
Thank you @antonbabenko!

PR #177, opened 2021-12-28
feat: Allow to pass additional trusted roles and services

mcanevet:
## Description

This change allows to add and additional trusted role and/or service that is allowed to assume the assumable role with OIDC.

## Motivation and Context

This module is often used in combination with IRSA. It is sometimes useful to be able to define not only a `sts:AssumeRoleWithWebIdentity`, but also a `sts:AssumeRole`. I have a IAM role per application which is assumed by the Service Account used by my pods (IRSA), but I also need another role to be allowed to use `sts:AssumeRole`, i.e. the role used by External Secrets (https://github.com/external-secrets/kubernetes-external-secrets#aws-ssm-parameter-store).

## How Has This Been Tested?


antonbabenko (commented):
Looks good, but please update an example for this submodule.

modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -60,6 +60,26 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
...
+    for_each = length(var.trusted_role_arns) + length(var.trusted_role_services) != 0 ? [true] : []
+
+    content {
+      effect = "Allow"
+
+      actions = ["sts:AssumeRole"]
+
+      principals {
+        type        = "AWS"
+        identifiers = var.trusted_role_arns
```
antonbabenko, line 74: Will it work well if `trusted_role_arns` or `trusted_role_services` is not specified?

modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -60,6 +60,26 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
       }
     }
   }
+
+  dynamic "statement" {
+    for_each = length(var.trusted_role_arns) + length(var.trusted_role_services) != 0 ? [true] : []
```
antonbabenko, line 65: Maybe split this into two separate statement blocks?

PR #184, opened 2022-02-02
feat: Add conditional policy statement attachments for EKS IAM role module

bryantbiggs:
## Description
- Adds new sub-module for IAM role for service accounts with conditional IAM policies for users to opt into various permissions for common EKS tasks. Initially, the following permissions have been provided and more can easily be added/integrated:
	- Cluster Autoscaler
	- External DNS
	- EBS CSI 
	- VPC CNI (both IPv4 and IPv6 are supported)
	- Node Termination Handler
	- Karpenter Controller - currently policy does not match 1:1  to what is prescribed; PR opened to align these https://github.com/aws/karpenter/pull/1332
- Miscellaneous updates to align standards
	- outputs changed to use `try()`
	- versions updated to newer, expanded syntax style

## Motivation and Context
- Closes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1825
- Closes https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1841

## Breaking Changes
- No

## How Has This Been Tested?


philicious:
@bryantbiggs thx for drafting this !

it definitely goes into the right direction.

from a `terraform-aws-modules/terraform-aws-eks` users perspective, it would probably be nice having this right in that module but I understand that you rather wanna put it into this iam-eks module. 
mentioning in docs or showing with examples how to easily integrate the two modules will help the user likewise

--

as for the question of which policies make sense to be included, I for myself would name:
- cluster-autoscaler (very common use case) 
- aws-loadbalancer-controller (mentioned in the EKS docs and afaiu now recommended to use over the in-tree version)
- AWS storage drivers as mentioned in AWS EKS docs: aws-ebs-csi-driver, aws-efs, aws-fsx. https://docs.aws.amazon.com/eks/latest/userguide/storage.html
Some rather uncommon but they are all in official AWS doc

max-rocket-internet:
@bryantbiggs happy to review but there's a million changes in this PR. Is that intentional??

modules/iam-assumable-role-with-saml/outputs.tf
```diff
@@ -1,19 +1,19 @@
 output "iam_role_arn" {
   description = "ARN of IAM role"
-  value       = element(concat(aws_iam_role.this.*.arn, [""]), 0)
+  value       = try(aws_iam_role.this[0].arn, "")
```
max-rocket-internet, line 3: Is this the official way of dealing with attribute outputs from resources that may or may not exist now? I like it, it's cleaner, but it works in all cases, like when on `destroy` also?

modules/iam-eks-role/policies.tf
```diff
@@ -0,0 +1,454 @@
...
+    actions = [
+      "autoscaling:DescribeAutoScalingGroups",
+      "autoscaling:DescribeAutoScalingInstances",
+      "autoscaling:DescribeLaunchConfigurations",
+      "autoscaling:SetDesiredCapacity",
+      "autoscaling:TerminateInstanceInAutoScalingGroup",
+      "ec2:DescribeInstanceTypes",
+    ]
+
+    resources = ["*"]
```
max-rocket-internet, line 21: This would be too broad for most people I think as it would allow termination of ANY EC2 instance in an ASG.

Personally I would copy the policy from [here](https://github.com/terraform-aws-modules/terraform-aws-eks/pull/716/files#diff-49c22f96d12f8b3391c9493b1ebcf2d0b8f9d6948fec4f0f3475338b0b1a7c6bR26-R66) as it includes the proper `condition` to limit the policy to this cluster. But then you would need to pass in the `cluster_id`...

max-rocket-internet: Could also be done in a later PR by someone who wants it

modules/iam-eks-role/policies.tf
```diff
@@ -0,0 +1,454 @@
...
+  policy      = data.aws_iam_policy_document.vpc_cni[0].json
+
+  tags = var.tags
+}
+
+resource "aws_iam_role_policy_attachment" "vpc_cni" {
+  count = var.create_role && var.attach_vpc_cni_policy ? 1 : 0
+
+  role       = aws_iam_role.this[0].name
+  policy_arn = aws_iam_policy.vpc_cni[0].arn
```
max-rocket-internet, line 349: You don't need to write this policy, it's already done by Amazon 😃

```suggestion
  policy_arn = "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy" 
```

modules/iam-eks-role/policies.tf
```diff
@@ -0,0 +1,454 @@
...
+
+resource "aws_iam_role_policy_attachment" "node_termination_handler" {
+  count = var.create_role && var.attach_node_termination_handler_policy ? 1 : 0
+
+  role       = aws_iam_role.this[0].name
+  policy_arn = aws_iam_policy.node_termination_handler[0].arn
+}
+
+################################################################################
+# Karpenter Controller Policy
```
max-rocket-internet, line 400: Totally up to you but I would strictly stick to policies for tools that are solely from AWS otherwise you will get PRs from people for every random service they want to use 😅

modules/iam-eks-role/policies.tf
bryantbiggs: That policy is only valid for IPv4 clusters; if you want to use IPv6 then you have to create your own https://github.com/bryantbiggs/terraform-aws-eks/blob/9c9ac81e3a3a91d52b5341d005de80a05a2fa0e8/node_groups.tf#L11

modules/iam-eks-role/policies.tf
max-rocket-internet: OK now I see that Karpenter is an AWS project. Never heard of it. So I guess ignore this comment!

modules/iam-eks-role/README.md
```diff
@@ -33,19 +38,54 @@ module "iam_eks_role" {
...
+      provider         = "oidc.eks.ap-southeast-1.amazonaws.com/id/5C54DDF35ER54476848E7333374FF09G"
+      provider_arn     = "arn:aws:iam::012345678901:oidc-provider/oidc.eks.ap-southeast-1.amazonaws.com/id/5C54DDF35ER54476848E7333374FF09G"
+      service_accounts = ["default:my-app-staging"]
+    }
   }
 }
 ```
 
-Note: the EKS clusters must in the current AWS region and account as they use the default AWS provider.
+This module has been design in conjunction with the [`terraform-aws-eks`](https://github.com/terraform-aws-modules/terraform-aws-eks) module and easily integrates with it:
```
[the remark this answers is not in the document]
bryantbiggs: likewise, we have made changes to make these more interoperable https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1870

modules/iam-eks-role/policies.tf
max-rocket-internet: OK cool, sounds good

modules/iam-eks-role/main.tf
```diff
@@ -1,58 +1,22 @@
...
-        ]
-      }
-
-      condition {
-        test     = "StringEquals"
-        variable = "${replace(data.aws_eks_cluster.main[statement.key].identity[0].oidc[0].issuer, "https://", "")}:sub"
-        values   = [for s in statement.value : "system:serviceaccount:${s}"]
-      }
-    }
-  }
```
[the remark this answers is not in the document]
bryantbiggs: unfortunately, this is the crux of the issue with this sub-module however. 
1. its not clear to users that the cluster name should be the key of the maps
2. it restricts users to creating a role in one region while roles are global - they will will have to figure out when to use `cluster_service_accounts` and when to use `provider_url_sa_pairs` when its not quite clear what the difference is on the surface
3. most importantly is the lifecycle issue. you cannot specify this role and a cluster and create at the same time. the cluster has to exist first

max-rocket-internet:
Maybe just ignore my previous comments about specifics and focus on the bigger picture with this PR and the `iam-eks-role` module: It looks like you want to use the `iam-eks-role` module to replace parts of the IAM policy code from the `terraform-aws-eks` module, and use it to apply preconfigured IAM policies for common "addons" like karpenter, external-dns etc. Is that correct? If so, I think that sounds great 👍

But then this functionality is also simply removing and replacing what I wrote this module for. It's now kind of an extension of the `terraform-aws-eks` module, making them closely related, based on the comments and examples in this PR. Also fine, but annoying for people who have complete separation between IRSA and EKS code and wanted a simple module to create many roles for use on EKS that are unrelated to cluster addons (like apps for example)

modules/iam-eks-role/policies.tf
bryantbiggs: ya, its the "new kid on the block"

modules/iam-eks-role/policies.tf
```diff
@@ -0,0 +1,454 @@
...
+# https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#iam-policy
+data "aws_iam_policy_document" "external_dns" {
+  count = var.create_role && var.attach_external_dns_policy ? 1 : 0
+
+  statement {
+    actions = ["route53:ChangeResourceRecordSets"]
+    resources = [
+      for hosted_zone in var.external_dns_hosted_zones :
+      "arn:${data.aws_partition.current.partition}:route53:::hostedzone/${hosted_zone}"
+    ]
```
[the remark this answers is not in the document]
bryantbiggs: good call out - fits with the other policies and corrected in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/184/commits/41d38278638f6ea2df00492b69deedfd5c0a304d

modules/iam-eks-role/policies.tf
bryantbiggs: thats a great point, skipped over that section in the docs. this too fits inline with the other policies where the docs are more open with `"*"` but here we are defaulting to open-ish while giving folks access to be more restrictive. added in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/184/commits/41d38278638f6ea2df00492b69deedfd5c0a304d

modules/iam-assumable-role-with-saml/outputs.tf
bryantbiggs: official - I don't know. however, I definitely use and abuse `try()` since it works so well in various use cases; outputs just happens to be one of those (and it shortens up the syntax a bit)

modules/iam-eks-role/main.tf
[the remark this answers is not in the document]
philicious: imho its not perfectly clear that the key of the map has to match the exact cluster name for functional reasons and isnt just meant to be an abbreviation to differentiate in the context of the map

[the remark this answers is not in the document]
bryantbiggs: To give some context, this change came about from 
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1825
- https://github.com/terraform-aws-modules/terraform-aws-eks/issues/1841

And I agree with the sentiments of those issues - these seem like common patterns many folks are repeating over and over and over (creating an IRSA role, specifying a policy, etc. - for common addons like VPC CNI, cluster autoscaler, etc.)

At first I thought something like this was being requested https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1827 but after clarifying, its really just the IRSA role w/ policy for common addons. Something where folks can just simply create an IRSA role and flip a flag to provide canned permissions for the use case of that role. 

So with that I came here because of the `iam-assumable-role-with-oidc` module, but then saw and remembered you added something for EKS and thought it would make the most sense to add on to this role to allow users to opt in to various canned permissions for common addons. However, in testing that I came across the lifecycle issues and  through testing cycles just reduced it down to the version you see now. 

Thats the context bit.

Regarding a path forward, I do believe we have to fix the lifecycle issue. Unfortunately this is how a lot of people work, especially when testing/trying out modules, plus we are seeing a lot more use cases where users are creating and destroying resources quite frequently in a CI pipeline type setup. So the targetted apply of `terraform apply -target module.vpc -target module.eks` is not going to fly. It also won't fly when we start propagating the use of this module to other modules and documentation. Its just one of the core tenants people rely on with Terraform - that Terraform will manage the lifecycle dependencies and not them.

Outside of that, personally I don't see the benefit in having the data source embedded with the restrictions it provides over having users provide the details. They can use the data source on their end if they like. And this is the area where I say that the module has been designed with in conjunction with the EKS module. I have aligned the inputs of this module with the outputs of that module so its very clear what is expected. That said, it doesn't lock people into using the EKS module; it just shows the relationship since I suspect this will most likely be the most used case.

max-rocket-internet:
> these changes aren't locking people into anything

No they are not but you are removing the clean implementation that was available in `iam-eks-role` so that it can work nicely with functionality for users of `terraform-aws-eks`. i.e. changing the docs to show implementation with `terraform-aws-eks`, specifically stating that it is now "design in conjunction with", removed the data resources etc.

> some assumptions (module will most likely be used with EKS and the EKS module in this organization

Exactly. The user that `iam-eks-role` was written for is like a user of EKS who has nothing to do with EKS infrastructure themselves. i.e. they just know the cluster name, `kubectl` or helm and don't need anything more. i.e. not a cluster administrator. For example a developer.

To give you context: we have about 50 EKS clusters managed by a "devops" team and about 400 developers who manage a whole bunch of apps across these clusters. In a large environment like this, EKS cluster TF resources can't be mixed in with application resources like in your docs, otherwise the TF state gets too big, blast radius etc. Developers need to create IRSA roles (we have 1000s of them) for their apps but have absolutely no knowledge of OIDC and very little of EKS. I didn't write the module for myself, I think there are others who want something clean and simple like this also. If you think outside of the bubble of EKS administration, there is a lot of general users, especially in larger orgs.

But anyway, do as you wish, I don't really mind. I just wanted to make you aware of why the module was written like that and who it was for 🙂

philicious:
as I spiked this addition in the eks-module and now seeing it being added here but having learned its origin from your convo, I can see both user-groups as stakeholders: those who want full flexibility and those who want some pre-canned policies.

shouldnt it be possible to keep the module flexible and satisfy both use-cases and not introduce breaking-changes ?

modules/iam-eks-role/main.tf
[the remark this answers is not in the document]
max-rocket-internet: All the context being given here is from the behemoth EKS module, used solely by EKS admins. Where the `iam-eks-role` module was written for regular users who have workloads on EKS who just want a simple IRSA role created.

max-rocket-internet:
> shouldnt it be possible to keep the module flexible and satisfy both use-cases and not introduce breaking-changes ?

Yes. Just don't delete the block of code that I commented on in my review. But TBH if we are going to merge the rest of the changes to the `iam-eks-role` module from this PR, then you might as well remove that code also because the module then becomes an extension of the EKS module. With the doc and example changes, all the boolean vars to add cluster component policies, then the `iam-eks-role` module will probably only be used by EKS admins.

Perhaps I should create a new module, called `iam-eks-role-simple` that explains better the use of the cluster name?

philicious:
I wouldn't argue it's not necessarily an extension of the EKS module. You could as well use it standalone and pass in the clustername or OIDC issuer url. No matter where it originates from. 


If you guys decide for keeping separate modules, it's probably more existing-user-friendly to keep this one as is and put the new stuff from this PR in a different and new module like "iam-eks-common" or such

bryantbiggs:
going with separate module for now I guess

philicious:
@bryantbiggs the implementation of the policy submodules looks really great so far 👍🏼 
Happy to see these modules coming 

How do you feel about also adding aws-load-balancer-controller policy ?
It's quite common also and by now preferable over the in-tree one 

Btw I will also give the policies a 3rd pair of eyes if you don't mind.

examples/iam-role-for-service-accounts-eks/main.tf
```diff
@@ -0,0 +1,223 @@
...
+  version = "~> 3.0"
+
+  name = local.name
+  cidr = "10.0.0.0/16"
+
+  azs             = ["${local.region}a", "${local.region}b", "${local.region}c"]
+  private_subnets = ["10.0.1.0/24", "10.0.2.0/24", "10.0.3.0/24"]
+  public_subnets  = ["10.0.4.0/24", "10.0.5.0/24", "10.0.6.0/24"]
+
+  enable_nat_gateway   = true
```
antonbabenko, line 187: If it is not critical for the example, can we disable NAT and VPC flow logs in the example to be able to spin up things faster&cheaper?

modules/iam-assumable-role-with-saml/outputs.tf
antonbabenko: @max-rocket-internet `try()` works better than using ternary operators for `destroy`. `try()` works in the same way as `element(concat(...))` but shorter.

examples/iam-role-for-service-accounts-eks/main.tf
```diff
@@ -0,0 +1,223 @@
...
+      provider                   = module.eks.oidc_provider
+      provider_arn               = module.eks.oidc_provider_arn
+      namespace_service_accounts = ["default:my-app", "canary:my-app"]
+    }
+  }
+
+  tags = local.tags
+}
+
+module "karpenter_controller_irsa_role" {
```
antonbabenko, line 152: Please add one `module` for a case where `create_role = false` because sometimes there can be some weird things we can't check in the happy path.

I don't remember the name of the module where this was a problem a few weeks/months ago.

antonbabenko (commented):
Looks very good. Minor comments here and there. :)

examples/iam-role-for-service-accounts-eks/main.tf
bryantbiggs: done!

modules/iam-role-for-service-accounts-eks/README.md
```diff
@@ -0,0 +1,174 @@
+# IAM Role for Service Accounts in EKS
```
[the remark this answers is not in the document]
bryantbiggs: Done - also ordered them to match the sub-module/example directory

bryantbiggs:
ok should be all set now @antonbabenko

examples/iam-role-for-service-accounts-eks/main.tf
bryantbiggs: > @bryantbiggs the implementation of the policy submodules looks really great so far 👍🏼 Happy to see these modules coming
> 
> How do you feel about also adding aws-load-balancer-controller policy ? It's quite common also and by now preferable over the in-tree one
> 
> Btw I will also give the policies a 3rd pair of eyes if you don't mind.

@philicious we'll add some more in a next PR, this one is getting big enough - hang tight

antonbabenko:
This PR is included in version 4.12.0 :tada:

examples/iam-role-for-service-accounts-eks/main.tf
antonbabenko: > done! (and it was Atlantis w/ EFS 😅 )

so it was so much recently :)

PR #185, opened 2022-02-02
feat: Update iam-read-only-policy - add default deny actions

Andrey9kin:
## Description
Added a variable to provide a list of IAM actions that will by denied. Default is set to s3:GetObject. See motivation and context for explanation why
 
## Motivation and Context
Based on the feedback from security community I want to add default deny actions since people can accidentally grant s3:GetObject and of objects are not encrypted with KMS CMK then S3 will be able to read object and return it to requester which might be undesired/can cause security issue

## Breaking Changes
N/A

## How Has This Been Tested?
Plus local tests

### DIFF
--- a/modules/iam-read-only-policy/variables.tf
+++ b/modules/iam-read-only-policy/variables.tf
@@ -27,6 +27,12 @@ variable "allowed_services" {
   type        = list(string)
 }
 
+variable "denied_actions" {
+  description = "List of actions to deny by default"
+  type        = list(string)
+  default     = ["s3:GetObject"]
+}
+
 variable "additional_policy_json" {
   description = "JSON policy document if you want to add custom actions"
   type        = string
--- a/examples/iam-read-only-policy/main.tf
+++ b/examples/iam-read-only-policy/main.tf
@@ -32,6 +32,7 @@ module "read_only_iam_policy" {
   description = "My read only example policy"
 
   allowed_services = local.allowed_services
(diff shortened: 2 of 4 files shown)


modules/iam-read-only-policy/variables.tf
```diff
@@ -27,6 +27,12 @@ variable "allowed_services" {
   type        = list(string)
 }
 
+variable "denied_actions" {
+  description = "List of actions to deny by default"
+  type        = list(string)
+  default     = ["s3:GetObject"]
```
antonbabenko, line 33:
```suggestion
  default     = []
```

It should be up to the user to decide what should be denied. Actually, `s3:GetObject` is pure read-only action :)

antonbabenko (changes requested):
Minor comment

modules/iam-read-only-policy/variables.tf
Andrey9kin: It is pure read indeed
This default is more of a guardrail since s3:GetObject gives to all resources will unintentionally give access to sensitive data. If everyone would use CMK's for the sensitive data according to security best practices then s3:GetObject would be not a problem because you would also need km:Decrypt for the key to get the object.

Anyhow, I see your point but got many people to point out that giving s3:GetObject is dangerous. So user's preference here is to disable it or replace it with more stuff.

Perhaps I should add this explanation to the description. What do you think?

PR #186, opened 2022-02-02
feat: Include cost explorer to default console services in `iam-read-only-policy` module

Andrey9kin:
## Description
Added Cost Explorer to default console services defaults.
 
## Motivation and Context
AWS updated home screen and now it shows data from cost explorer thus generating Cloud Trail accessed denied events if policy does not container Cost Explorer

## Breaking Changes
N/A

## How Has This Been Tested?
Plus local tests


antonbabenko:
This PR is included in version 4.11.0 :tada:

Issue #193, opened 2022-02-19
iam-eks-role cannot attach a policy created in the same module using role_policy_arns input


SilverXXX:
## Description
When using iam-eks-role i cannot pass a role_policy_arn i created in the same module

## Versions
 - Terraform: 1.1.4
 - Provider(s):
   - registry.terraform.io/hashicorp/aws: 3.74.0
 - Module:  
   - terraform-aws-modules/eks/aws: 18.2.7
   - terraform-aws-modules/iam/aws//modules/iam-eks-role: 4.13.1

## Reproduction
Using the code snippet below and 

### Code Snippet to Reproduce
```
resource "aws_iam_policy" "alb_load_balancer" {
    name        = "K8SALBController"
    path        = local.base_iam_path
    description = "Policy that allows k8s load balancer controller to provisione alb/elb"
    policy = file("${path.module}/policies/alb-policy.json")         
}

module "aws_alb_controller_role" {
  source = "terraform-aws-modules/iam/aws//modules/iam-eks-role"
  //source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
  version = "4.13.1"

  create_role      = true
  role_name = "aws-alb-controller"
  role_path = local.base_iam_path
  role_description = "IRSA role for load balancer controller"

  role_policy_arns               = [aws_iam_policy.alb_load_balancer.arn]
  cluster_service_accounts = {
    "${var.cluster_name}" = [
      "kube-system:aws-alb-controller"
    ]
  }
  depends_on = [
    module.eks.cluster_id
  ]
}
```

## Expected behavior
Policy should be attached

## Actual behavior
I get the error below

### Terminal Output
```
│ Error: Invalid for_each argument
│
│   on .terraform\modules\eks.aws_node_termination_handler_role\modules\iam-eks-role\main.tf line 76, in resource "aws_iam_role_policy_attachment" "custom":
│   76:   for_each = toset([for arn in var.role_policy_arns : arn if var.create_role])
│     ├────────────────
│     │ var.create_role is true
│     │ var.role_policy_arns is list of string with 1 element
│
│ The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work around this, use the -target argument to first apply only the resources that the for_each depends on.
```
## Additional context
Using
```
resource "aws_iam_role_policy_attachment" "custom" {
  role       = module.aws_alb_controller_role.iam_role_name
[...]

PR #197, opened 2022-03-03
fix: Use count instead of for_each in iam-eks-role (#193)

petur:
## Description
This is the same method as is used in iam-assumable-role-with-oidc. Since the argument to count is a list, not a set, the number of elements doesn't depend on the values, and terraform can decide how many elements are needed before creating the policies.

## Motivation and Context
This fixes https://github.com/terraform-aws-modules/terraform-aws-iam/issues/193

This issue is a blocker for using this module in automation.

## Breaking Changes
This changes the address of the `aws_iam_role_policy_attachment` resource, since it's index by a number instead of a string.
This will cause the resource to be recreated when applying the module after changes.

There are no changes to the outputs of the module, and the created resources should be identical.

## How Has This Been Tested?

This was also tested in our automated tests, which create a role and a policy to attach to a service account, like so:

```
resource "aws_iam_policy" "foo" {
  name   = "${var.cluster_name}-foo"
  policy = data.aws_iam_policy_document.foo.json
}

module "foo_role" {
  source  = "terraform-aws-modules/iam/aws//modules/iam-eks-role"
  version = "4.13.1"

  role_description = "IRSA role for foo"
  role_name_prefix = "foo-"
  role_policy_arns = [aws_iam_policy.foo.arn]

  cluster_service_accounts = {
    (var.cluster_name) = ["${var.namespace}:${local.service_account_name}"]
  }
}
```

The change has no effects on the code that uses the module - except that it works when applying in an empty workspace.

### DIFF
--- a/modules/iam-eks-role/main.tf
+++ b/modules/iam-eks-role/main.tf
@@ -73,8 +73,8 @@ resource "aws_iam_role" "this" {
 }
 
 resource "aws_iam_role_policy_attachment" "custom" {
-  for_each = toset([for arn in var.role_policy_arns : arn if var.create_role])
+  count = var.create_role ? length(var.role_policy_arns) : 0
 
   role       = aws_iam_role.this[0].name
-  policy_arn = each.key
+  policy_arn = var.role_policy_arns[count.index]
 }


lorengordon:
Please see [my comment on the issue](https://github.com/terraform-aws-modules/terraform-aws-iam/issues/193#issuecomment-1063421930) for a better option...

adminomargallo (approved):

bryantbiggs:
I don't think we will be going back to count, but instead going with the great suggestion by @lorengordon - however, we have to batch these changes in the next breaking change for the module (we have some other issues around provider version and such and we'll probably update and align those to be more current)

antonbabenko:
This issue has been resolved in version 5.0.0 :tada:

PR #199, opened 2022-03-04
feat: Add load_balancer_controller targetgroup binding only role

cdhesse:
## Description
This implements a least privilege role for load_balancer_controller when using TargetGroup binding only as documented [here](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/deploy/installation/#setup-iam-manually).

## Motivation and Context
The current module has a role for load_balancer_controller, but it is the full set of permissions, which are not required when implementing TargetGroup binding only.

## Breaking Changes
No

## How Has This Been Tested?

- [x]  I have tested and validated these changes using one or more of the provided `examples/*` projects

I tested the changes by executing the examples/iam-role-for-service-accounts-eks, and added a module call there to create the new role.


cdhesse:
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2542 was also created to address the missing describevps iam permission.

cdhesse:
> [kubernetes-sigs/aws-load-balancer-controller#2542](https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/2542) was also created to address the missing describevps iam permission.

Issue 2542 has been closed as of today in the aws-load-balancer-controller repo.

modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -740,3 +740,47 @@ resource "aws_iam_role_policy_attachment" "load_balancer_controller" {
...
+  count = var.create_role && var.attach_load_balancer_controller_targetgroup_binding_only_policy ? 1 : 0
+
+  statement {
+    actions = [
+      "ec2:DescribeSecurityGroups",
+      "ec2:DescribeInstances",
+      "ec2:DescribeVpcs",
+      "elasticloadbalancing:DescribeTargetGroups",
+      "elasticloadbalancing:DescribeTargetHealth",
+      "elasticloadbalancing:ModifyTargetGroup",
```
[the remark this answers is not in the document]
cdhesse: The use case for TargetGroupBinding only mode, vs. full ALB, is that you want to create (or pre-existing) TargetGroups outside of the kubernetes environment.  This change would require the TargetGroup in question to have the elbv2.k8s.aws/cluster tag.  While on one hand this solution ensures that at least the owner/creator of the TargetGroup intended for it to be updated from a load-balancer-controller, the downside is there may be valid reasons for that tag to not exist.

The second thought - your proposed change is a better version than what is documented as part of the load-balancer-controller intro guide.   A person trying for the first time to use this module would quickly realize it didn't work, until they reverse engineer by looking at this policy that they need to go back and tag their TargetGroups with this k8s tag.  I suggest that if you implement it this way, the ideal would be that the TargetBingingOnly documentation should include this version of the role vs. their more permissive one.  Then the requirement to tag your TargetGroups would seem a more appropriate place.

examples/iam-role-for-service-accounts-eks/main.tf
```diff
@@ -182,6 +182,22 @@ module "load_balancer_controller_irsa_role" {
   tags = local.tags
 }
 
+module "load_balancer_controller_targetgroup_binding_only_irsa_role" {
+  source = "../../modules/iam-role-for-service-accounts-eks"
+
+  role_name                                                       = "load_balancer_controller"
```
bryantbiggs, line 188:
```suggestion
  role_name                                                       = "load_balancer_controller_targetgroup_binding_only"
```

bryantbiggs: what a name 😅

modules/iam-role-for-service-accounts-eks/policies.tf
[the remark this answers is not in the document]
bryantbiggs: For now we'll proceed with what is defined by the upstream documentation. We can revisit later for any additional policy scoping

modules/iam-role-for-service-accounts-eks/variables.tf
```diff
@@ -177,3 +177,11 @@ variable "attach_load_balancer_controller_policy" {
   type        = bool
   default     = false
 }
+
+# https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/targetgroupbinding/targetgroupbinding/#reference
+# https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/deploy/installation/#setup-iam-manually
+variable "attach_load_balancer_controller_targetgroup_binding_only_policy" {
+  description = "Determines whether to attach the Load Balancer Controller policy for the TargetGroup binding only scenario."
```
bryantbiggs, line 184:
```suggestion
  description = "Determines whether to attach the Load Balancer Controller policy for the TargetGroupBinding only"
```

bryantbiggs:
@cdhesse if you could make the requested tweaks, update your branch with the latest, and re-run `pre-commit run -a`, we should be set to merge this then. Thank you for the PR!

cdhesse:
> @cdhesse if you could make the requested tweaks, update your branch with the latest, and re-run `pre-commit run -a`, we should be set to merge this then. Thank you for the PR!

I've done as you requested, should be ready to merge now, thank you!

bryantbiggs (approved):
thanks again for the PR @cdhesse - good to go @antonbabenko 👍🏽

antonbabenko:
This PR is included in version 4.16.0 :tada:

PR #212, opened 2022-03-29
fix: Backwards compatibility in 4.x.x series in iam-user submodule

emyller:
## Description
Revert AWS provider version requirement for recent release, in order to prevent something like the following error out of a `terraform init`:

```
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider hashicorp/aws:
│ no available releases match the given constraints >= 2.45.0, ~> 3.0, < 4.0.0,
│ >= 4.1.0
```

## Motivation and Context
Not break compatibility in workspaces that still couldn't upgrade their AWS provider -- but probably should eventually.

## Breaking Changes
The opposite, actually.

## How Has This Been Tested?
`terraform init` working again.

### DIFF
--- a/modules/iam-user/README.md
+++ b/modules/iam-user/README.md
@@ -24,13 +24,13 @@ This module outputs commands and PGP messages which can be decrypted either usin
 | Name | Version |
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | >= 0.12.6 |
-| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 4.1.0 |
+| <a name="requirement_aws"></a> [aws](#requirement\_aws) | >= 2.50 |
 
 ## Providers
 
 | Name | Version |
 |------|---------|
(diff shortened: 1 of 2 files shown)


antonbabenko:
Thanks for this PR. Yes, there is a small mess with the versions (my apologies!). I think your fix is good enough for now and we will bump the version of the whole module at some point later.

antonbabenko:
This PR is included in version 4.17.1 :tada:

ferrandinand:
Looks like password attribute does not exist regarding to the oficial documentation in provider version 3.X
https://registry.terraform.io/providers/erindatkinson/aws-test/latest/docs/resources/iam_user_login_profile

```
  on .terraform/modules/iam/modules/iam-user/outputs.tf line 33, in output "iam_user_login_profile_password":
  33:   value       = try(aws_iam_user_login_profile.this[0].password, "")
This object has no argument, nested block, or exported attribute named
"password".
```

Issue #226, opened 2022-04-15
policies.tf should be arn partition consistent


xchan:
## Description

All aws resource should consider partition.

- [*] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]:

- Terraform version:
v1.1.8
- Provider version(s):
4.20.1

## Reproduction Code [Required]

https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/karpenter/main.tf

terraform apply

## Expected behavior

The example should work on region with parition such as aws-gov and aws-cn

## Actual behavior

Karpenter failed to install due to malformated arn document

### Terminal Output Screenshot(s)

```shell
Error: error creating IAM Policy AmazonEKS_Karpenter_Controller_Policy-2022041412360163110000000d: MalformedPolicyDocument: Partition "aws" is not valid for resource "arn:aws:ec2::xxx:subnet/".
│ status code: 400, request id: d5d4e215-5abb-49d2-a0fe-dee528de002d
│
│ with module.infra_cluster.module.karpenter_irsa.aws_iam_policy.karpenter_controller[0],
│ on .terraform/modules/infra_cluster.karpenter_irsa/modules/iam-role-for-service-accounts-eks/policies.tf line 584, in resource "aws_iam_policy" "karpenter_controller":
│ 584: resource "aws_iam_policy" "karpenter_controller" {
```

## Additional context

```code
diff --git a/modules/iam-role-for-service-accounts-eks/policies.tf b/modules/iam-role-for-service-accounts-eks/policies.tf
index 16b09e7..570d43b 100644
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -17,7 +17,7 @@ data "aws_iam_policy_document" "cert_manager" {

   statement {
     actions   = ["route53:GetChange"]
-    resources = ["arn:aws:route53:::change/*"]
+    resources = ["arn:${local.partition}:route53:::change/*"]
   }

   statement {
[...]

PR #229, opened 2022-04-18
fix: AWS arn partition fix for Karpenter

xchan:
## Description

This fix will replace all hardcoded aws arn to be parition targed

## Motivation and Context

https://github.com/terraform-aws-modules/terraform-aws-iam/issues/226

## Breaking Changes

N/A

## How Has This Been Tested?

- [*] I have executed `pre-commit run -a` on my pull request

I have successfully launched Karpenter on "aws-cn" partiton. This fix should also make "aws-gov" work as well

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -17,7 +17,7 @@ data "aws_iam_policy_document" "cert_manager" {
 
   statement {
     actions   = ["route53:GetChange"]
-    resources = ["arn:aws:route53:::change/*"]
+    resources = ["arn:${local.partition}:route53:::change/*"]
   }
 
   statement {
@@ -26,7 +26,7 @@ data "aws_iam_policy_document" "cert_manager" {
       "route53:ListResourceRecordSets"
     ]
 
-    resources = var.cert_manager_hosted_zone_arns
+    resources = [for s in var.cert_manager_hosted_zone_arns : replace(s, "arn:aws:", "arn:${local.partition}:")]
   }
 
   statement {
@@ -263,7 +263,7 @@ data "aws_iam_policy_document" "ebs_csi" {
         "kms:RevokeGrant",
       ]
 
-      resources = var.ebs_csi_kms_cmk_ids
+      resources = [for s in var.ebs_csi_kms_cmk_ids : replace(s, "arn:aws:", "arn:${local.partition}:")]
 
       condition {
         test     = "Bool"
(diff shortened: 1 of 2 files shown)


modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -26,7 +26,7 @@ data "aws_iam_policy_document" "cert_manager" {
       "route53:ListResourceRecordSets"
     ]
 
-    resources = var.cert_manager_hosted_zone_arns
+    resources = [for s in var.cert_manager_hosted_zone_arns : replace(s, "arn:aws:", "arn:${local.partition}:")]
```
bryantbiggs, line 29: we should leave this as is - its up to users to provide the correct ARNs

modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -263,7 +263,7 @@ data "aws_iam_policy_document" "ebs_csi" {
         "kms:RevokeGrant",
       ]
 
-      resources = var.ebs_csi_kms_cmk_ids
+      resources = [for s in var.ebs_csi_kms_cmk_ids : replace(s, "arn:aws:", "arn:${local.partition}:")]
```
bryantbiggs, line 266: same as above, leave as is

modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -284,7 +284,7 @@ data "aws_iam_policy_document" "ebs_csi" {
         "kms:DescribeKey",
       ]
 
-      resources = var.ebs_csi_kms_cmk_ids
+      resources = [for s in var.ebs_csi_kms_cmk_ids : replace(s, "arn:aws:", "arn:${local.partition}:")]
```
bryantbiggs, line 287: same for all of these - we don't modify the variables users provide us, they should provide the correct ARNs

modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -987,7 +989,7 @@ resource "aws_iam_role_policy_attachment" "node_termination_handler" {
 data "aws_iam_policy_document" "vpc_cni" {
   count = var.create_role && var.attach_vpc_cni_policy ? 1 : 0
 
-  # arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy
+  # arn:${local.partition}:iam::aws:policy/AmazonEKS_CNI_Policy
```
bryantbiggs, line 992: this is just a comment showing where the policy came from so leave as is

xchan:
Agreed.
The inline arn should be arn awared with leaving variable arn what it is.
After this MR
https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/karpenter/main.tf
need to be specify var.karpenter_controller_ssm_parameter_arns for local.partition arn

I will update this MR later.

PR #236, opened 2022-04-22
feat: Improved iam-eks-role module (simplified, removed provider_url_sa_pairs, updated docs)

### LINKED ISSUE #219: Explain the differences between iam-role-for-service-accounts-eks and iam-eks-role modules
## Is your request related to a new offering from AWS?
No

It seems like there are duplicate modules with [iam-role-for-service-accounts-eks](https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks) and [iam-eks-role](https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-eks-role).

Documentation needs to be more clear on which route users should take or better yet deprecate one of the modules as there is a lot of overlap.

n/a
## Additional context
n/a

max-rocket-internet:
## Description

There was some misunderstanding in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/184 about the intended use of the `iam-eks-role` module. I think a lot of it can be put down to that there was an assumption that this module was in some way related to the `terraform-aws-eks` module, but it is not. So I thought I would come back just to make a few minor adjustments to clear up the confusion 🙂

Fixes #219

The main points were:

From @bryantbiggs:

> 1. its not clear to users that the cluster name should be the key of the maps
> 2. it restricts users to creating a role in one region while roles are global
> 3. most importantly is the lifecycle issue. you cannot specify this role and a cluster and create at the same time. the cluster has to exist first
> 4. This module has been design in conjunction with the [`terraform-aws-eks`]

From @philicious:
> imho its not perfectly clear that the key of the map has to match the exact cluster name

So here I am:

- Clarifying docs for the above points
- Remove `provider_url_sa_pairs` option: it's a very advanced option IMO and not for the intended use of this module. Removing it makes it simpler. If people want this option then they can just use [iam-assumable-role-with-oidc](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-assumable-role-with-oidc/)
- Adjust outputs to use nicer `try`
- Provide 3 clear examples

## Motivation and Context

Just to clarify and better document the module.

## Breaking Changes

Yes if someone was using `provider_url_sa_pairs`.

## How Has This Been Tested?

- [x] I have tested it in my environment also


antonbabenko:
LGTM, but I wonder what @bryantbiggs thinks about this as he has added another EKS-related module and probably has a better picture of what is what.

max-rocket-internet:
> but I wonder what @bryantbiggs thinks about this as he has added another EKS-related module and probably has a better picture of what is what.

The 2 modules are unrelated. The module he added was specifically for cluster admins, this module is for anyone 🙂

antonbabenko:
This PR is included in version 4.23.0 :tada:

PR #243, opened 2022-05-03
feat: Add `policy_name_prefix` for IRSA policies

### LINKED ISSUE #239: Feature Request: Support prefix on policy name
## Is your request related to a new offering from AWS?

No

We have specific naming convention requirements for roles and policies which is preventing us from being able to adopt this module. Role name prefix was added previously https://github.com/terraform-aws-modules/terraform-aws-iam/pull/101 but a similar feature is not available for Policy name prefix.

I may be able to contribute this back, but confirm there was buy-in in the first place to add it (along with the preferred approach, see two variations listed below).

Ideally a new (optional) variable would be added for a policy name prefix to either completely override the current value of `name_prefix` in `aws_iam_policy` or just provide a value which is pre-pended to the default

For example either of the following would be fine:

```terraform
resource "aws_iam_policy" "cluster_autoscaler" {
  count = var.create_role && var.attach_cluster_autoscaler_policy ? 1 : 0

  name_prefix = "${var.policy_name_prefix}AmazonEKS_Cluster_Autoscaler_Policy-"
...

}
```

or

```terraform
resource "aws_iam_policy" "cluster_autoscaler" {
  count = var.create_role && var.attach_cluster_autoscaler_policy ? 1 : 0

  name_prefix = var.policy_name_prefix
...

}
```

It's unclear to me the viability of the second option, in the case where the role needs to attach multiple policies the consumer would only be able to specify a single prefix and may lose some of the niceties of having something in the policy name which describes what the policy is for. As of now, i think this module only attaches one policy per role so that may be a non-issue in the short term.

The only available option right now would be to fork the module and exposing this configuration

## Additional context

N/A

cvockrodt:
## Description
Replace `AmazonEKS_` with user provided variable `policy_name_prefix` which defaults to `AmazonEKS_`

## Motivation and Context

Many organizations have specific naming convention requirements for roles and policies. Allowing for name prefixes means members of those organizations can utilize this module and be compliant. 

Closes #239 

## Breaking Changes

No breaking changes.

## How Has This Been Tested?


bryantbiggs (approved):
Thanks for the PR @cvockrodt

antonbabenko:
This PR is included in version 4.24.0 :tada:

PR #257, opened 2022-06-09
feat: Allow access to keypairs for Karpenter controller policy

pavlospt:
When Karpenter provisioners are using a custom `LaunchTemplate` there is a chance that it includes a custom key-pair.

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -571,6 +571,7 @@ data "aws_iam_policy_document" "karpenter_controller" {
       "arn:${local.partition}:ec2:*:${local.account_id}:volume/*",
       "arn:${local.partition}:ec2:*:${local.account_id}:network-interface/*",
       "arn:${local.partition}:ec2:*:${coalesce(var.karpenter_subnet_account_id, local.account_id)}:subnet/*",
+      "arn:${local.partition}:ec2:*:${local.account_id}:key-pair/*",
     ]
   }
 


bryantbiggs:
I'm not sure about this addition since Karpenter relies on the AmazonSSMManagedInstanceCore policy and therefore remote access is available via SSM

pavlospt:
> I'm not sure about this addition since Karpenter relies on the AmazonSSMManagedInstanceCore policy and therefore remote access is available via SSM

Since adding a Keypair is a feature of Launch templates why would someone limit this? We usually connect on our instances via SSH if we need to debug something! I mean the fact that there is remote connection through SSM Managed Instance is kind of irrelevant to the fact that one can use a custom keypair!

bryantbiggs:
Understood - but this is not a recommended practice over using something like SSM where credentials are ephemeral and access can be controlled through policies. 

I think if you want to add this permission, the route I would suggest today is to add it through a custom policy and attach it to the role using `role_policy_arns`.

pavlospt:
> Understood - but this is not a recommended practice over using something like SSM where credentials are ephemeral and access can be controlled through policies.
> 
> I think if you want to add this permission, the route I would suggest today is to add it through a custom policy and attach it to the role using `role_policy_arns`.

Hello @bryantbiggs , thank you for your feedback! 

This is what I am currently doing in order to work around it, but I am not sure why this might not be a feature that could be added. Even if it is not a recommended practice, I certainly believe there are a lot customers out there that will probably migrate from CA to Karpenter and may be using LTs with custom keypairs. 🤔

jimbocoder:
Identical issue here, was going to make the same PR. It may not be the default but I could suggest making it an optional flag? Arriving at the solution of attaching an additional policy takes some time to hunt down the real issue and come to that conclusion; this difficulty will be multiplied by the number of users migrating from CA to Karpenter using this module, so it might be worth making more explicit at least.

pavlospt:
> Identical issue here, was going to make the same PR. It may not be the default but I could suggest making it an optional flag? Arriving at the solution of attaching an additional policy takes some time to hunt down the real issue and come to that conclusion; this difficulty will be multiplied by the number of users migrating from CA to Karpenter using this module, so it might be worth making more explicit at least.

Indeed, this is my main reasoning as well, of course I am open to changing things to optional if needed, or adding a flag to create an additional policy attachment and not have it included by default!

pavlospt:
This PR shouldn’t close IMO!

pavlospt:
This PR still shouldn’t close IMO!

dindurthy:
> Even if it is not a recommended practice, I certainly believe there are a lot customers out there that will probably migrate from CA to Karpenter and may be using LTs with custom keypairs. 🤔

We are exactly in this category.

pavlospt:
Nope!

Issue #282, opened 2022-09-28
self_assume_role var.role_name is null


FernandoMiguel:
## Description

Please provide a clear and concise description of the issue you are encountering, and a reproduction of your configuration (see the `examples/*` directory for references that you can copy+paste and tailor to match your configs if you are unable to copy your exact configuration). The reproduction MUST be executable by running `terraform init && terraform apply` without any further changes.

If your request is for a new feature, please use the `Feature request` template.

- [X] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: v5.5

- Terraform version:
terraform -v
Terraform v1.2.9
on darwin_arm64
+ provider registry.terraform.io/gavinbunney/kubectl v1.14.0
+ provider registry.terraform.io/hashicorp/aws v4.32.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.2.0
+ provider registry.terraform.io/hashicorp/helm v2.6.0
+ provider registry.terraform.io/hashicorp/http v3.1.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.13.1
+ provider registry.terraform.io/hashicorp/null v3.1.1
+ provider registry.terraform.io/hashicorp/random v3.4.3
+ provider registry.terraform.io/hashicorp/time v0.8.0
+ provider registry.terraform.io/hashicorp/vault v3.8.2
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1

## Reproduction Code [Required]
```
module "argocd_iam_role_sa_cross_cluster" {
  source  = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts-eks"
  version = "~> v5.5"

  role_name_prefix       = substr("${var.addon_context.eks_cluster_id}-argocd-sts-", 0, 38)
  max_session_duration   = "3600"
  force_detach_policies  = true
  allow_self_assume_role = true

  oidc_providers = {
    argocd = {
      provider_arn = var.addon_context.eks_oidc_provider_arn
      namespace_service_accounts = [
        "${var.namespace}:argocd-application-controller",
        "${var.namespace}:argocd-server",
      ]
    }
  }
}
```

## Expected behavior

for a role to be created

## Actual behavior

```
│ Error: Invalid template interpolation value
│
[...]

PR #283, opened 2022-09-28
fix: Explicitly assume with condition matching role arn

FernandoMiguel:
Signed-off-by: Fernando Miguel <github@FernandoMiguel.net>

## Description
Fix the self assume role using a condition 

## Motivation and Context
The policy can not use the role arn in the principal field until the role already exists.
And since we can't create a role without a trust policy, we have chicken-egg problem.

Using a condition to match the role arn instead of using it in the principal, we are able to work around the problem.

## How Has This Been Tested?

tested this against one of our cluster, using my own fork branch

fixes: #282

### DIFF
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -6,6 +6,7 @@ locals {
     replace(url, "https://", "")
   ]
   number_of_role_policy_arns = coalesce(var.number_of_role_policy_arns, length(var.role_policy_arns))
+  role_name_condition        = var.role_name != null ? var.role_name : "${var.role_name_prefix}*"
 }
 
 data "aws_caller_identity" "current" {}
@@ -31,7 +32,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
       condition {
         test     = "ArnLike"
         variable = "aws:PrincipalArn"
-        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${var.role_name}"]
+        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${var.role_name_condition}"]
       }
     }
   }
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -25,7 +25,13 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
 
       principals {
         type        = "AWS"
-        identifiers = ["arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:role${var.role_path}${var.role_name}"]
+        identifiers = ["*"]
+      }
+
+      condition {
+        test     = "ArnLike"
+        variable = "aws:PrincipalArn"
+        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${var.role_name}"]
       }
     }
   }
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -32,7 +32,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
       condition {
         test     = "ArnLike"
         variable = "aws:PrincipalArn"
-        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${var.role_name_condition}"]
+        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${local.role_name_condition}"]
       }
     }
   }
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/antonbabenko/pre-commit-terraform
-    rev: v1.75.0
+    rev: v1.76.0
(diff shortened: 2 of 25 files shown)


bryantbiggs:
if I understand the issue linked - this is still not valid, no? If a user uses `role_name_prefix`, even the changes in this PR do not address that scenario, correct?

FernandoMiguel:
@bryantbiggs there is no way to use `role_name_prefix` cause that is only known after role being created.
in this PR i am only addressing the fact that the policy was not valid, which is now, making this module usable in our use-case with argocd.

I still think we should default `role_path` to `"/"` which is the entirely most common scenario.
and for `role_name_prefix` maybe a condition in the var if `allow_self_assume_role` is present

FernandoMiguel:
this is a sample of the plan output
```hcl
  # module.base_system.module.argocd[0].module.argocd_iam_role_sa_cross_cluster.data.aws_iam_policy_document.this[0] will be read during apply
  # (config refers to values not yet known)
 <= data "aws_iam_policy_document" "this" {
      + id   = (known after apply)
      + json = (known after apply)

      + statement {
          + actions = [
              + "sts:AssumeRole",
            ]
          + effect  = "Allow"
          + sid     = "ExplicitSelfRoleAssumption"

          + condition {
              + test     = "ArnLike"
              + values   = [
                  + (known after apply),
                ]
              + variable = "aws:PrincipalArn"
            }

          + principals {
              + identifiers = [
                  + "*",
                ]
              + type        = "AWS"
            }
        }
      + statement {
          + actions = [
              + "sts:AssumeRoleWithWebIdentity",
            ]
          + effect  = "Allow"

          + condition {
              + test     = "StringEquals"
              + values   = [
                  + "sts.amazonaws.com",
                ]
              + variable = "oidc.eks.us-east-1.amazonaws.com/id/XXXX:aud"
            }
          + condition {
              + test     = "StringEquals"
              + values   = [
                  + "system:serviceaccount:argocd-system:argocd-application-controller",
                  + "system:serviceaccount:argocd-system:argocd-server",
                ]
              + variable = "oidc.eks.us-east-1.amazonaws.com/id/XXXX:sub"
            }

          + principals {
              + identifiers = [
                  + "arn:aws:iam::XXXX:oidc-provider/oidc.eks.us-east-1.amazonaws.com/id/XXXX",
                ]
              + type        = "Federated"
            }
        }
    }
```

bryantbiggs:
so a few things here:
1. The whole premise for using the manually constructed ARN of `"arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:role${var.role_path}${var.role_name}"` is because of this chicken vs the egg scenario. We can't reference the IAM role ARN in the assume role policy because the role needs the policy but the policy needs the role. So we "cheat" in these scenarios and re-construct what the result will be so this should be valid today for users who use just `role_name` because its all static data known at plan/apply time
2. The role path most likely needs to be updated. The provider allows for `null` which is interpreted as `"/"` but with our re-construction of the ARN we have to be explicit and set the default as `"/"` and can't use the `null` trick
3. We can, and should, address the `role_name_prefix` here. Simply doing something like a `coalesce(var.role_name, var.role_name_prefix)` would suffice in the ArnLike changes here (I think). Or we can have two values, one for `var.role_name` and one for `var.role_name_prefix` which might need to end with `*` to match correctly

FernandoMiguel:
Using a wildcard in the case of prefix might work, did not test that.
I can try to code the changes Friday, as I'm mostly offline tomorrow.

I entirely agree with #2

FernandoMiguel:
@bryantbiggs I've tried to address your concerns

FernandoMiguel:
HI team,
Can this be reviewed?

Thanks in advance.

modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -25,7 +26,13 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
...
       principals {
         type        = "AWS"
-        identifiers = ["arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:role${var.role_path}${var.role_name}"]
+        identifiers = ["*"]
+      }
+
+      condition {
+        test     = "ArnLike"
+        variable = "aws:PrincipalArn"
+        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${var.role_name_condition}"]
```
bryantbiggs, line 35: please see the failed CI checks - these should all be a local
```suggestion
        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${local.role_name_condition}"]
```

FernandoMiguel:
I've added some code to handle 3 roles. 
but can't really support `name_prefix` on those, without it being a breaking change.

FernandoMiguel:
Team, pls review.
Thanks

bryantbiggs:
still have some pre-commit checks to fix but overall I think it looks good - @antonbabenko thoughts?

FernandoMiguel:
@bryantbiggs `pre-commit` ran, and docs updated

FernandoMiguel:
Hello folks, 
Can this PR be review?
Thanks in advance.

antonbabenko (approved):

antonbabenko:
@FernandoMiguel Thank you very much for this PR!

antonbabenko:
This PR is included in version 5.5.2 :tada:

PR #288, opened 2022-10-14
feat: Add support for wildcards to service account defs for iam-eks-role

endzyme:
## Description
Added support for wildcards in service account conditions, to more closely align with similar support on `iam-assumable-role-with-oidc` submodule. I've also updated the documentation with `terraform-docs` as noted in the pre-commit hooks.

## Motivation and Context

Allows people to enable many service accounts in an EKS cluster to assume this role without explicitly adding each service account name to the trust policy. This is also backwards compatible and follows a similar variable conventions which exists in another submodule in this repository.

## Breaking Changes

None. Recommending a minor version bump due to an additional variable, but has an empty default.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-eks-role/README.md
+++ b/modules/iam-eks-role/README.md
@@ -108,6 +108,7 @@ No modules.
 |------|-------------|------|---------|:--------:|
 | <a name="input_allow_self_assume_role"></a> [allow\_self\_assume\_role](#input\_allow\_self\_assume\_role) | Determines whether to allow the role to be [assume itself](https://aws.amazon.com/blogs/security/announcing-an-update-to-iam-role-trust-policy-behavior/) | `bool` | `false` | no |
 | <a name="input_cluster_service_accounts"></a> [cluster\_service\_accounts](#input\_cluster\_service\_accounts) | EKS cluster and k8s ServiceAccount pairs. Each EKS cluster can have multiple k8s ServiceAccount. See README for details | `map(list(string))` | `{}` | no |
+| <a name="input_cluster_service_accounts_with_wildcards"></a> [cluster\_service\_accounts\_with\_wildcards](#input\_cluster\_service\_accounts\_with\_wildcards) | EKS cluster and k8s ServiceAccount pairs. Each EKS cluster can have multiple k8s ServiceAccount. See README for details | `map(list(string))` | `{}` | no |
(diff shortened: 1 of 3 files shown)


endzyme:
Upon testing with the provided examples I've hit an annoying Terraformism: 

```local.eks_cluster_names is a set of dynamic, known only after apply```

I will try to clear this up for the example.

endzyme:
Hi there - anything else I can do here to move this along?

bryantbiggs:
I believe this is already supported in https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks

endzyme:
> I believe this is already supported in https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks

Thanks for letting me know. Is there a reason to have two modules so similar in intent in this repo? I am having trouble determining when to use one over the other.

Also is your comment to let me know that since it's already in the other module, it would not be a valuable addition to this module?

bryantbiggs:
>Thanks for letting me know. Is there a reason to have two modules so similar in intent in this repo? I am having trouble determining when to use one over the other.

It was an error on our part and to remove it now would be a breaking change for users. [https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks?rgh-link-date=2022-11-26T04%3A42%3A21Z should ](https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks) is the role users should utilize which offers the most flexibility for these various use cases and configurations

endzyme:
> > Thanks for letting me know. Is there a reason to have two modules so similar in intent in this repo? I am having trouble determining when to use one over the other.
> 
> It was an error on our part and to remove it now would be a breaking change for users. https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks is the role users should utilize which offers the most flexibility for these various use cases and configurations

Is there any way to indicate this on the submodule? It would be really good to know if you're going to get rid of it because it was a mistake. Maybe even just mentioning it in a header that people _should_ be using the other module instead.

In fact the docs currently outline how the iam-eks-role module differentiates itself from the iam-role-for-service-accounts-eks module and other submodules in this same repo.

It would probably go a long way to people spending time on contributions to help move the project forward.

Issue #291, opened 2022-10-17
IRSA module missing policy when using only  `attach_load_balancer_controller_targetgroup_binding_only_policy = true`


apamildner:
## Description
When using the option `attach_load_balancer_controller_targetgroup_binding_only_policy = true`, a subset of the
policies are added that by the name of the argument should allow targetgroup binding to occur. This may have been true before, but as of AWS Load Balancer Controller `v2.4.4` installed via chart version `1.4.5` [here](https://artifacthub.io/packages/helm/aws/aws-load-balancer-controller),  the policies are not enough any more. We also need these to be added:

```
                "ec2:AuthorizeSecurityGroupIngress",
                "ec2:RevokeSecurityGroupIngress"
```
More specifically, here is where they need to be added:
[here](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts-eks/policies.tf#L870)

## How was this discovered
As described, deploy the AWS Load Balancer controller and then add a TargetGroupBinding.
Reading the logs from AWS Load Balancer controller pods we see there are authorization issues related to the above mentioned policies missing.
**TGB applied**
```yaml
apiVersion: elbv2.k8s.aws/v1beta1
kind: TargetGroupBinding
metadata:
  namespace: networking
  name: traefik-service-external-http
spec:
  serviceRef:
    name: traefik-service-external
    port: 80
  targetGroupARN: <arn>
  targetType: instance
  networking:
    ingress:
      - from:
        - ipBlock:
            cidr: 0.0.0.0/0
        ports:
        - port: 30002 
          protocol: TCP
```
## Workaround
```hcl
module "load_balancer_controller" {
  source                                                          = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts-eks"
  version                                                         = "~> v5.5.0"
  role_name                                                       = "aws-load-balancer-controller-${var.cluster_name}"
  attach_load_balancer_controller_targetgroup_binding_only_policy = true
  oidc_providers = {
    main = {
      provider_arn               = module.eks.oidc_provider_arn
      namespace_service_accounts = ["kube-system:aws-load-balancer-controller"]
    }
  }
}
# Temporary fix
resource "aws_iam_policy" "auth_revoke_sg_policy" {
  name        = "AuthorizeRevokeIngressSG"
  description = "Required when using targetGroupBinding inside EKS"

  policy = <<EOF
{
    "Statement": [
        {
[...]

PR #292, opened 2022-10-17
fix: Added additional policies required for targetgroup binding only to work 

apamildner:
Added additional policies required for targetgroup binding only to work with AWS LBC

## Description
Adds two additional policies to the subset of policies that are in effect when only using TargetGroupBinding policies with the AWS Load Balancer Controller.

## Motivation and Context
The policies applied when using only `var.attach_load_balancer_controller_targetgroup_binding_only_policy = true`
is not sufficient for target group binding to work in latest version of AWS LBC. Not sure when the issue arose, and it may be related to how I define inline ingresses. Nonetheless I am only using TargetGroupBinding resources and the policy fails for that use case. More details described here: [issue](https://github.com/terraform-aws-modules/terraform-aws-iam/issues/291)

## Breaking Changes
This doesn't break backwards compatibility

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -863,12 +863,14 @@ data "aws_iam_policy_document" "load_balancer_controller_targetgroup_only" {
       "ec2:DescribeSecurityGroups",
       "ec2:DescribeInstances",
       "ec2:DescribeVpcs",
+      "ec2:AuthorizeSecurityGroupIngress",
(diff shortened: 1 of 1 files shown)


bryantbiggs (approved):
Thank you for the detailed findings and fix @apamildner 🎉

apamildner:
@bryantbiggs Thanks for quick approval. Can we release this? I think maybe the chore bot is not picking it up due to the commit name.

antonbabenko:
This PR is included in version 5.5.3 :tada:

PR #301, opened 2022-11-04
fix: Wrong content for oidc_fully_qualified_audiences in iam-assumable-role-with-oidc

corneliusgdahling:
Fix: fully qualified audiences should generate StringEquals but is generating StringLike

## Description

There is a bug that causes oidc_fully_qualified_audiences to create StringLike condition instead of StringEquals condition. This is simply because the content is created with StringLike instead of StringEquals

## Motivation and Context

Solves bug

## Breaking Changes

I don't know

## How Has This Been Tested?

### DIFF
--- a/examples/iam-assumable-role-with-oidc/main.tf
+++ b/examples/iam-assumable-role-with-oidc/main.tf
@@ -50,3 +50,22 @@ module "iam_assumable_role_self_assume" {
 
   oidc_fully_qualified_subjects = ["system:serviceaccount:default:sa1", "system:serviceaccount:default:sa2"]
 }
+
+#####################################
+# IAM assumable role with audience
+#####################################
+module "iam_assumable_role_with_oidc" {
+  source = "../../modules/iam-assumable-role-with-oidc"
+
+  create_role = true
+
+  role_name = "role-with-oidc-github-actions"
+
+  provider_url = "token.actions.githubusercontent.com"
+
+  role_policy_arns = [
+    "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+  ]
+
+  oidc_fully_qualified_audiences = ["sts.amazonaws.com"]
+}
--- a/examples/iam-assumable-role-with-oidc/main.tf
+++ b/examples/iam-assumable-role-with-oidc/main.tf
@@ -67,5 +67,6 @@ module "iam_assumable_role_with_oidc" {
     "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
   ]
 
+  assume_role_condition_test     = "StringEquals"
   oidc_fully_qualified_audiences = ["sts.amazonaws.com"]
 }
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -75,7 +75,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
         for_each = length(var.oidc_fully_qualified_audiences) > 0 ? local.urls : []
 
         content {
-          test     = "StringEquals"
+          test     = var.assume_role_condition_test
           variable = "${statement.value}:aud"
           values   = var.oidc_fully_qualified_audiences
         }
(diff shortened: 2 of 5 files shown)


modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -75,7 +75,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
         for_each = length(var.oidc_fully_qualified_audiences) > 0 ? local.urls : []
 
         content {
-          test     = "StringLike"
+          test     = "StringEquals"
```
bryantbiggs, line 78: I think this is better suited to be a variable if we need to change it - see https://github.com/terraform-aws-modules/terraform-aws-iam/blob/3ec0f0fd5d360019231f14fc7431418b52deda0c/modules/iam-role-for-service-accounts-eks/variables.tf#L76

corneliusgdahling: Fixed

examples/iam-assumable-role-with-oidc/main.tf
```diff
@@ -50,3 +50,22 @@ module "iam_assumable_role_self_assume" {
 
   oidc_fully_qualified_subjects = ["system:serviceaccount:default:sa1", "system:serviceaccount:default:sa2"]
 }
+
+#####################################
+# IAM assumable role with audience
```
bryantbiggs, line 55: This example doesn't match the changes being made

corneliusgdahling: Should be fixed now

modules/iam-assumable-role-with-oidc/variables.tf
```diff
@@ -105,3 +105,9 @@ variable "allow_self_assume_role" {
   type        = bool
   default     = false
 }
+
+variable "assume_role_condition_test" {
+  description = "Name of the [IAM condition operator](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html) to evaluate when assuming the role"
+  type        = string
+  default     = "StringEquals"
```
bryantbiggs, line 112:
```suggestion
  default     = "StringLike"
```

examples/iam-assumable-role-with-oidc/main.tf
```diff
@@ -50,3 +50,23 @@ module "iam_assumable_role_self_assume" {
...
+  source = "../../modules/iam-assumable-role-with-oidc"
+
+  create_role = true
+
+  role_name = "role-with-oidc-github-actions"
+
+  provider_url = "token.actions.githubusercontent.com"
+
+  role_policy_arns = [
+    "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
```
bryantbiggs, line 67:
```suggestion
    "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess",
```

examples/iam-assumable-role-with-oidc/main.tf
```diff
@@ -50,3 +50,23 @@ module "iam_assumable_role_self_assume" {
 
   oidc_fully_qualified_subjects = ["system:serviceaccount:default:sa1", "system:serviceaccount:default:sa2"]
 }
+
+#####################################
+# IAM assumable role with audience
+#####################################
+module "iam_assumable_role_with_oidc" {
```
bryantbiggs, line 57:
```suggestion
module "iam_assumable_role_condition_test" {
```

bryantbiggs:
based on your example - is this change needed or are you looking for https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-github-oidc-role ?

corneliusgdahling:
> based on your example - is this change needed or are you looking for https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-github-oidc-role ?

Kind of. That would only work for GitHub, while making this change would allow it to work with others such as CircleCI

corneliusgdahling:
> > based on your example - is this change needed or are you looking for https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-github-oidc-role ?
> 
> Kind of. That would only work for GitHub, while making this change would allow it to work with others such as CircleCI

I have long since created my own module with this implementation, but still see the value in this. What do you think @bryantbiggs ?

PR #308, opened 2022-11-16
feat: Add support for creating IAM GitHub OIDC provider and role(s)

bryantbiggs:
## Description
- Add support for creating IAM GitHub OIDC provider and role(s)

## Motivation and Context
- The two additional sub-modules allow users to setup an IAM GitHub OIDC provider plus roles that will trust this provider. This is commonly used in things like GitHub action workflows without the need for static credentials

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/modules/iam-github-oidc-role/main.tf
+++ b/modules/iam-github-oidc-role/main.tf
@@ -1,0 +1,73 @@
+data "aws_partition" "current" {}
+data "aws_caller_identity" "current" {}
+
+locals {
+  github_token_url = "token.actions.githubusercontent.com"
+
+  account_id = data.aws_caller_identity.current.account_id
+  partition  = data.aws_partition.current.partition
+  dns_suffix = data.aws_partition.current.dns_suffix
+}
+
+################################################################################
+# GitHub OIDC Role
+################################################################################
+
+data "aws_iam_policy_document" "this" {
+  count = var.create ? 1 : 0
+
+  statement {
+    sid    = "GithubOidcAuth"
+    effect = "Allow"
+    actions = [
+      "sts:TagSession",
+      "sts:AssumeRoleWithWebIdentity"
+    ]
+
+    principals {
+      type        = "Federated"
+      identifiers = ["arn:${local.partition}:iam::${local.account_id}:oidc-provider/${local.github_token_url}"]
+    }
+
+    condition {
+      test     = "ForAllValues:StringEquals"
+      variable = "${local.github_token_url}:iss"
+      values   = ["sts.${local.dns_suffix}"]
+    }
+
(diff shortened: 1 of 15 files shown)


modules/iam-github-oidc-role/main.tf
```diff
@@ -0,0 +1,73 @@
+data "aws_partition" "current" {}
+data "aws_caller_identity" "current" {}
+
+locals {
+  github_token_url = "token.actions.githubusercontent.com"
```
antonbabenko, line 5: I think this ok for the main github.com, but it will have to be parametrized when using with GitHub Enterprise, as described [here](https://sjramblings.io/github-actions-aws-authentication-with-oidc-for-github-enterprise/) (search for `mygithub.com/_services/token`).

Let's make this as a variable with this value as the default.

antonbabenko (changes requested):
Minor comment for GitHub Enterprise users, LGTM.

modules/iam-github-oidc-role/main.tf
bryantbiggs: good callout, however - I don't know if that post is correct. For example, it does not appear to reference the AWS STS service (secure token service) anywhere so I'm a little confused. I've reached out internally to see if I can get in contact with someone at GitHub for some guidance. I'll update once I know a bit more

antonbabenko: 👍 Maybe the post I found was not the most accurate one.

bryantbiggs: eh, I don't know if I know enough to weigh in on whether its correct or not 😅 - since its security related, I figured its better to error on the side of caution and see if I can find someone who would be able to weigh in on the matter

antonbabenko: Currently, you've made it as a variable (with the correct default value), so it should be good enough for merge.

bryantbiggs: well, that post doesn't have some things like this https://github.com/terraform-aws-modules/terraform-aws-iam/pull/308/files#diff-490344238465f9b048f0ff98d48bfe068162248472dba85614109d31edf6f269R33-R37

which made me realize they don't have any reference to the secure token service which I believe is a hard requirement for these two entities to create a trust relationship (between GitHub and AWS)

antonbabenko: Right. Hopefully, someone with GHE experience can verify and confirm it.

antonbabenko (approved):
Looks great 👍

antonbabenko:
This PR is included in version 5.6.0 :tada:

PR #321, opened 2022-12-19
feat: Add `create_custom_role_trust_policy` to control when a `custom_role_trust_policy` should be used

### LINKED ISSUE #320: Module "iam-assumable-role" Invalid count value when policy is computed
## Description

Let's suppose the following definition of a role
```
module "integration-role" {
  source  = "terraform-aws-modules/iam/aws//modules/iam-assumable-role"
  version = "~> 4.3"

  custom_role_trust_policy = jsonencode({
    Version = "2012-10-17"
    Statement = concat([
      {
        Effect = "Allow"
        Principal = {
          AWS = [
            "arn:aws:iam::${data.aws_caller_identity.this.account_id}:root",
          ]
        }
        Action = [
          "sts:AssumeRole",
        ],
      }],
      # This is going to add a list of statements with diff external IDs (computed)
      [
        for name in var.stack_names :
          data.spacelift_aws_integration_attachment_external_id.this[name].assume_role_policy_statement
    ])
  })

  create_role = true

  role_name         = var.name
  role_requires_mfa = false

  custom_role_policy_arns = var.role_policy_arns
}

```
Since `custom_role_trust_policy` is used as conditional within the resource with a `count`, the following error is shown
```
╷
│ Error: Invalid count argument
│
│   on .terraform/modules/integration.integration-role/modules/iam-assumable-role/main.tf line 6, in data "aws_iam_policy_document" "assume_role":
│    6:   count = var.custom_role_trust_policy == "" && var.role_requires_mfa ? 0 : 1
│
│ The "count" value depends on resource attributes that cannot be determined
│ until apply, so Terraform cannot predict how many instances will be
│ created. To work around this, use the -target argument to first apply only
│ the resources that the count depends on.
╵
╷
│ Error: Invalid count argument
│
│   on .terraform/modules/integration.integration-role/modules/iam-assumable-role/main.tf line 35, in data "aws_iam_policy_document" "assume_role_with_mfa":
│   35:   count = var.custom_role_trust_policy == "" && var.role_requires_mfa ? 1 : 0
│
│ The "count" value depends on resource attributes that cannot be determined
│ until apply, so Terraform cannot predict how many instances will be
│ created. To work around this, use the -target argument to first apply only
│ the resources that the count depends on.
```

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: 5.9.2

- Terraform version:
1.3.6
- Provider version(s):
Terraform v1.3.6
on darwin_arm64
+ provider registry.terraform.io/hashicorp/aws v4.47.0
+ provider registry.terraform.io/spacelift-io/spacelift v0.1.35

Steps to reproduce the behaviour:
- Define a role using a computed element as part of the `custom_role_trust_policy`
as shown above in the description.
- Run a plan

## Expected behaviour

The plan is to be shown without errors.

## Actual behaviour

The error shown above in the description is shown.

### Terminal Output Screenshot(s)

Already on description

## Additional context

Notice I'm trying to create external IDs with a diff provider, but the same can be reproduced with any computed value.

egarbi:
## Description
Fixes #320 

## Motivation and Context

This is needed to prevent errors when trying to add a  computed value for `custom_role_trust_policy` 

## Breaking Changes

Those already passing valid values into `custom_role_trust_policy` will need to 
add an extra input `create_custom_role_trust_policy = true` to prevent Terraform
from trying to destroy policies.

## How Has This Been Tested?


egarbi:
@antonbabenko can you please take a look at this? I proposed a PR too

modules/iam-assumable-role/variables.tf
```diff
@@ -83,11 +83,20 @@ variable "custom_role_policy_arns" {
...
 
 variable "custom_role_trust_policy" {
-  description = "A custom role trust policy"
+  description = "A custom role trust policy. (Only valid if create_custom_role_trust_policy = true)"
   type        = string
   default     = ""
 }
 
+variable "create_custom_role_trust_policy" {
+  description = <<EOF
```
antonbabenko, line 92: Please write it in one line like "Whether to create a custom role trust policy".

modules/iam-assumable-role/main.tf
```diff
@@ -9,7 +9,7 @@ locals {
 }
 
 data "aws_iam_policy_document" "assume_role" {
-  count = var.custom_role_trust_policy == "" && var.role_requires_mfa ? 0 : 1
+  count = var.create_custom_role_trust_policy && var.role_requires_mfa ? 0 : 1
```
antonbabenko, line 12: The condition should be reverted to be the same as before. Now, `create_custom_role_trust_policy` is `false`, by default. Please verify.

antonbabenko (changes requested):

modules/iam-assumable-role/variables.tf
egarbi: 4be6b92547033b41acfc231cf18658ba866bf1cb

modules/iam-assumable-role/main.tf
egarbi: I don't really understand your point here, `create_custom_role_trust_policy = false` and `custom_role_trust_policy == ""` are equivalent. right?
Although, I found an issue happening when  `create_custom_role_trust_policy` was false and custom_role_trust_policy was not empty. I introduced a fix here: 1d9c0e43283ee6e22f637e18ec7fb4236e99b31c

egarbi: Done 84a47eacda09cb35960310bfe6c634f1b770ddc2

egarbi:
@antonbabenko would you mind checking again, please

schniber:
Hey all,

@egarbi: I think you need to mark the changes that @antonbabenko  requested as done in order to move forward with the process.

@bryantbiggs: any guidance on how to help get this merged ?

Thanks a lot !

Bests.

bryantbiggs (approved):

bryantbiggs:
@antonbabenko ok from your end?

antonbabenko (approved):

antonbabenko:
This PR is included in version 5.30.0 :tada:

fregatte:
Hi mates, small note here from my side.
This is a breaking change because we are required to make changes in your code to keep the behavior the same as the previous version.
Let me explain. I have configured `custom_role_trust_policy` with the `5.29.2` module and by default, it was created on apply. On the `5.30.0` version default behavior was changed to `create=false` and to have the same behavior I need to add the `create_custom_role_trust_policy = true` explicitly. It's not a minor change in my opinion and it can break something in case you don't use strict constraints, e.g. `~>5.0`.
So, #420 above is an example of such an issue.
Do you agree or do I have a lack of understanding of modules versioning?

egarbi:
> Hi mates, small note here from my side. This is a breaking change because we are required to make changes in your code to keep the behavior the same as the previous version. Let me explain. I have configured `custom_role_trust_policy` with the `5.29.2` module and by default, it was created on apply. On the `5.30.0` version default behavior was changed to `create=false` and to have the same behavior I need to add the `create_custom_role_trust_policy = true` explicitly. It's not a minor change in my opinion and it can break something in case you don't use strict constraints, e.g. `~>5.0`. So, #420 above is an example of such an issue. Do you agree or do I have a lack of understanding of modules versioning?

In my defence, I will say that the breaking change was mentioned in the description.

fregatte:
> In my defence, I will say that the breaking change was mentioned in the description.

I'm not trying to blame so you no need to defend :)
Just highlighted that IMO it would be more appropriate to bump the major version of the module instead of the minor. Just to clarify my understanding and make a note for future releases.

egarbi:
> > In my defence, I will say that the breaking change was mentioned in the description.
> 
> I'm not trying to blame so you no need to defend :) Just highlighted that IMO it would be more appropriate to bump the major version of the module instead of the minor. Just to clarify my understanding and make a note for future releases.

I agree

Issue #325, opened 2023-01-12
modules/iam-role-for-service-accounts-eks - wrong condition string in IAM policy for EBS driver


prysmakou-sa:
## Description

The condition string for DeleteVolume action has changed in the [example policy of the aws-ebs-csi-driver ](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/ce8a68ba1d2ad5b142edccc52ee09fa82a097965/docs/example-iam-policy.json#L104) and in the (AWS managed) AmazonEBSCSIDriverPolicy.

It must be:
```
{
      "Effect": "Allow",
      "Action": [
        "ec2:DeleteVolume"
      ],
      "Resource": "*",
      "Condition": {
        "StringLike": {
          "ec2:ResourceTag/kubernetes.io/created-for/pvc/name": "*"
        }
      }
    },

```
compare it to[ the current](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/aa7ddba39b658088788652b1a0c3221ccf73cf13/modules/iam-role-for-service-accounts-eks/policies.tf#L223):
```
statement {
    actions   = ["ec2:DeleteVolume"]
    resources = ["*"]

    condition {
      test     = "StringLike"
      variable = "ec2:ResourceTag/kubernetes.io/cluster/*"
      values   = ["owned"]
    }
  }
```
EKS 1.23

aws-ebs-csi-driver-2.11.1 chart (the driver version 1.11.3)

## Steps to reproduce
create PVC following https://github.com/kubernetes-sigs/aws-ebs-csi-driver/tree/master/examples/kubernetes/dynamic-provisioning

Delete the resources

## Expected behavior

Volume deletion without errors

## Actual behavior

EBS controller logs like "DeleteDisk could not delete volume: UnauthorizedOperation: You are not authorized to perform this operation. " 

Example tags:
<img width="807" alt="Screenshot 2023-01-12 at 13 51 54" src="https://user-images.githubusercontent.com/81765132/212074368-814aa664-13a4-4f85-94ae-d04cb9172274.png">

prysmakou-sa:
For the sake of truth I'm able to reproduce the error only on some of my clusters. But the policy in this repo is definitely different from what we have in AWS managed policy and from what we can find in example policy in aws-ebs-csi-driver repo (they also use it for testing purposes)

bryantbiggs:
[...]

PR #326, opened 2023-01-16
fix: Reflect the changes in the ebs_csi driver

prysmakou-sa:
## Description

modules/iam-role-for-service-accounts-eks -
EBS-CSI driver policy - fix condition string for DeleteVolume action to reflect the changes in the upstream

## Motivation and Context
Issue - https://github.com/terraform-aws-modules/terraform-aws-iam/issues/325
The change in the upstream - https://github.com/kubernetes-sigs/aws-ebs-csi-driver/pull/1450

## Breaking Changes
There are no breaking changes

## How Has This Been Tested?
- tested on the EKS 1.23, Chart: aws-ebs-csi-driver-2.11.1, the driver version: 1.11.3

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -226,6 +226,17 @@ data "aws_iam_policy_document" "ebs_csi" {
   }
 
   statement {
+    actions   = ["ec2:DeleteVolume"]
+    resources = ["*"]
+
+    condition {
+      test     = "StringLike"
+      variable = "ec2:ResourceTag/kubernetes.io/created-for/pvc/name"
+      values   = ["*"]
+    }
+  }
+
+  statement {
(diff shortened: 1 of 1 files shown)


bryantbiggs:
so the change looks great - makes sense. However, for users who are using versions < 2.14.1, this will break their permissions - *I think. 

So we'll need to somehow factor that into the change - I don't know if the CSI drivers still support the old route with the cluster owned, open to ideas though

prysmakou-sa:
@bryantbiggs just to be on the safe side we could let loose the permissions for a while: to save the current/former statement and add the new one. What do you think?

prysmakou-sa:
I have updated the PR so now we do not replace the condition but add the new statement.

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.11.1 :tada:

PR #328, opened 2023-01-17
docs: Add example of custom IRSA role

ricosega:
## Description
The current module is very useful but it only provides solutions to ones of the most used things, cert-manager, external-secrets, etc.
But there are no options for any other permissions. The idea is to take profit of this module and allow people to use it in order to add custom policies other than the ones that are already set.

## Motivation and Context
In my case I needed that some of my workloads were able to assume a role with a serviceaccount which has permissions to write in a specific bucket.

## How Has This Been Tested?

I tested it with the solution provided in the examples/ (see files changed)

### DIFF
--- a/.github/workflows/lock.yml
+++ b/.github/workflows/lock.yml
@@ -8,7 +8,7 @@ jobs:
   lock:
     runs-on: ubuntu-latest
     steps:
-      - uses: dessant/lock-threads@v3
+      - uses: dessant/lock-threads@v4
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           issue-comment: >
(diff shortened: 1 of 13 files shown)


bryantbiggs:
Isn't this suported today with the `var.role_policy_arns` variable? users can bring their own policy to use with the role

ricosega:
> Isn't this suported today with the `var.role_policy_arns` variable? users can bring their own policy to use with the role

you are right, as there are so many options I just did not see it.
what do you think about only updating the readme and example like this?

bryantbiggs:
sure, we can update the readme to make this more clear

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.14.0 :tada:

PR #313, opened 2022-11-22
feat: Allow multiple MFA devices and users to manage MFA devices

rballan:
## Description

Fix the `IAMSelfManagement` policy to allow virtual MFA devices creation.

Due to [this new feature](https://aws.amazon.com/blogs/security/you-can-now-assign-multiple-mfa-devices-in-iam/), the resource name of virtual MFA devices change and the policy doesn't work anymore. 

The wildcard in the MFA resource name is required now, because the user must specify a name for his virtual MFA device, and this is a part of the resource name.

## Motivation and Context
- Resolves #312 

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

- [x] I have directly edited the `IAMSelfManagement` policy and added successfully a new virtual MFA device.

PR #340, opened 2023-02-15
Allow Change Password when no MFA present

0x46616c6b:
We faced an issue since #313 that new users which need to set a new password after first login not able to do that. Reason was that `iam:ChangePassword` was not granted for users which not have a MFA device set (which is the case for new users).

With this change we are able to create new users and they can login with there temporary password and are able to reset the password.


antonbabenko (approved):
Looks good to me

antonbabenko:
This PR is included in version 5.11.2 :tada:

PR #355, opened 2023-03-23
Output IAM Role Assume Role Policy JSON

irasnyd:
## Description

For each submodule that creates a single IAM Role, I added a Terraform output containing the Assume Role Policy JSON.

## Motivation and Context

This will allow users to combine the Assume Role Policy from multiple instances of these Terraform modules, then attach the combined policy to a single role.

My use case is to create an IAM Role that can be used by Kubernetes Pods (EKS IRSA) as well as be assumed by another IAM Role. [This Terraform documentation](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#example-of-merging-source-documents) shows how to merge multiple policies together.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-assumable-role-with-oidc/README.md
+++ b/modules/iam-assumable-role-with-oidc/README.md
@@ -62,6 +62,7 @@ No modules.
 | Name | Description |
 |------|-------------|
 | <a name="output_iam_role_arn"></a> [iam\_role\_arn](#output\_iam\_role\_arn) | ARN of IAM role |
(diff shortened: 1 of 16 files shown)


bryantbiggs:
Thank you for the PR. However, we won't be going forward with this change. There are two primary issues here:
1. Removing the count conditional is a breaking change
2. We already have an IRSA module today https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks

PR #357, opened 2023-03-27
fix: Remove `"autoscaling:UpdateAutoScalingGroup"` permission from cluster-autoscaler IRSA

blanchardma:
## Description
Remove unnecessary action ("autoscaling:UpdateAutoScalingGroup") from the IAM policy of the Cluster Autoscaler.
Update the order of the IAM actions as in the [documentation,](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#full-cluster-autoscaler-features-policy-recommended) so that it is easier for future comparison.

## Motivation and Context
Reduces the IAM permissions to the required actions.

## Breaking Changes
No.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -74,8 +74,7 @@ data "aws_iam_policy_document" "cluster_autoscaler" {
     content {
       actions = [
         "autoscaling:SetDesiredCapacity",
-        "autoscaling:TerminateInstanceInAutoScalingGroup",
-        "autoscaling:UpdateAutoScalingGroup",
+        "autoscaling:TerminateInstanceInAutoScalingGroup"
       ]
 
       resources = ["*"]


modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -59,11 +59,8 @@ data "aws_iam_policy_document" "cluster_autoscaler" {
       "autoscaling:DescribeLaunchConfigurations",
       "autoscaling:DescribeScalingActivities",
       "autoscaling:DescribeTags",
-      "ec2:DescribeLaunchTemplateVersions",
```
bryantbiggs, line 62: we can remove `"autoscaling:UpdateAutoScalingGroup"` since its not listed in the cluster-autoscaler policy on the project, but the rest of the changes need to be reverted. The statement below has a condition which many of these permissions will fail on

bryantbiggs (changes requested):

modules/iam-role-for-service-accounts-eks/policies.tf
blanchardma: First thanks you for the quick response. Based on the related [documentation](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#full-cluster-autoscaler-features-policy-recommended) it is possible to use it with condition and tags like here: _"In this example, only the second block of actions should be updated to restrict the resources/add conditionals:"_ And also the EC2 instances has the tag from autoscaling group. For "eks:DescribeNodegroup" in the 2nd [example](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md#minimal-iam-permissions-policy) is it similar with ARN but not same. Below the example the API use case with "eks:DescribeNodegroup" is described but I could not test so far.

blanchardma: @bryantbiggs I can adjust the PR, but I wanted to ask if the additional information has it described a little better and impact  first feedback?

bryantbiggs: I'm sorry but I am not following. from what I can see from the upstream project, we should only be removing `autoscaling:UpdateAutoScalingGroup` and not modifying anything else in this PR

blanchardma: @bryantbiggs the PR is update as discussed.

modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -74,8 +74,7 @@ data "aws_iam_policy_document" "cluster_autoscaler" {
     content {
       actions = [
         "autoscaling:SetDesiredCapacity",
-        "autoscaling:TerminateInstanceInAutoScalingGroup",
-        "autoscaling:UpdateAutoScalingGroup",
+        "autoscaling:TerminateInstanceInAutoScalingGroup"
```
bryantbiggs, line 77:
```suggestion
        "autoscaling:TerminateInstanceInAutoScalingGroup",
```

blanchardma: Now it's fixed. 👍

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.17.1 :tada:

PR #370, opened 2023-04-10
feat: Add support for other remote PGP keys in iam-user module

jasondamour:
## Description
Use http datasource to retrieve keys from non-keybase urls (like keys.openpgp.org)

## Motivation and Context
Allows alternatives to keybase, without needing to hardcode pgp keys

## How Has This Been Tested?

### DIFF
--- a/modules/iam-user/main.tf
+++ b/modules/iam-user/main.tf
@@ -19,7 +19,7 @@ resource "aws_iam_user_login_profile" "this" {
   count = var.create_user && var.create_iam_user_login_profile ? 1 : 0
 
   user                    = aws_iam_user.this[0].name
-  pgp_key                 = startswith(var.pgp_key, "http") ? data.http.pgp_key.response_body : var.pgp_key
+  pgp_key                 = startswith(var.pgp_key, "http") ? data.http.pgp_key[0].response_body : var.pgp_key
   password_length         = var.password_length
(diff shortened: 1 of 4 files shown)


modules/iam-user/main.tf
```diff
@@ -13,7 +19,7 @@ resource "aws_iam_user_login_profile" "this" {
   count = var.create_user && var.create_iam_user_login_profile ? 1 : 0
 
   user                    = aws_iam_user.this[0].name
-  pgp_key                 = var.pgp_key
+  pgp_key                 = startswith(var.pgp_key, "http") ? data.http.pgp_key[0].response_body : var.pgp_key
```
antonbabenko, line 22:
```suggestion
  pgp_key                 = try(data.http.pgp_key[0].response_body, var.pgp_key)
```

modules/iam-user/main.tf
```diff
@@ -1,3 +1,9 @@
+data "http" "pgp_key" {
+  count           = startswith(var.pgp_key, "http") ? 1 : 0
```
antonbabenko, line 2: Please add an empty line after `count`.

antonbabenko (changes requested):
Please update `versions.tf` to include `http` provider block and update the docs.

PR #349, opened 2023-03-15
fix: Add `kms:decrypt` policy for External Secret

zhuoyang:
## Description
Add `kms:decrypt` policy for External Secret

## Motivation and Context
`kms:decrypt` policy is needed to get secrets from secret manager.

Reference https://github.com/external-secrets/external-secrets/issues/390

## Breaking Changes
N/A

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [x] I have executed `pre-commit run -a` on my pull request

PR #375, opened 2023-04-21
fix(security): Don't decrypt all KMS keys by default

dev-slatto:
## Description

As of #349 a pretty large security vulnerability was introduced to the `iam-role-for-service-accounts-eks` module that by default allows full decryption off all KMS keys within the account that it's implemented in. In the release notes it was noted as a fix and not a new feature. For the naked eye this can indicate that it's a fix in some existing feature that they might not be using, when the reality is that it opened up a backdoor to all KMS keys if you use the External Secrets role.

## Motivation and Context

I've changed the default here to be an empty array to focus on closing this security vulnerability first. I'll try to make some time to implement the condition object to the `kms:decrypt` statement later on. 

## Breaking Changes

This might be a breaking changes for those that have already implemented the security vulnerability and should be noted as such in the release documentation. 

## How Has This Been Tested?


dev-slatto:
Hey @antonbabenko @bryantbiggs! 👋🏻 
Could you guys have a look at this as soon as possible as it might resolve quite a large vulnerability? 😄 

As of now it dosen't resolve the issue to set the variable to an empty list or string as the statement block isn't dynamic and therefore generates a IAM policy statement that's not valid. A work around is to put in a random KeyId that's not in your account.

bryantbiggs:
We can switch to a dynamic block to all users to disable, but the default value should stay. Also, this is not a security vulnerability - it may be more permission than what's necessary but that's it

dev-slatto:
One could agree with that, but with keeping the default value if violates [AWS Key Management Service Best
Practices](https://d1.awsstatic.com/whitepapers/aws-kms-best-practices.pdf) (see section regarding `Least Privilege / Separation of Duties`).

If you still want to keep the default value I can change that one back in this PR, but it should be mentioned somewhere that by keeping the default value your might put your account at risk.

PR #378, opened 2023-05-11
feat: Add support for AWS Gateway controller (VPC Lattice) to IRSA module

bryantbiggs:
## Description
- Add support for AWS Gateway controller (VPC Lattice) to IRSA module
- Add new variables `cluster_autoscaler_cluster_names` and `karpenter_controller_cluster_name` which shadow their current equivalents (`cluster_autoscaler_cluster_ids` and `karpenter_controller_cluster_id`). Users should now use the `cluster_name` output from the EKS module (starting in v19.x) since `cluster_id` now has a different meaning on local clusters on Outposts. This change starts to correct that behavior and align the inputs/outputs to reduce confusion but without breaking compatibility. The `_id` equivalents will be removed at the next breaking change
- Updated the VPC and EKS module versions to the latest major versions; example has been updated and validated to reflect these changes

## Motivation and Context
- IRSA role and policy support for AWS Gateway controller - this is used for VPC Lattice integration

## Breaking Changes
- No

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/main.tf
+++ b/modules/iam-role-for-service-accounts-eks/main.tf
@@ -7,7 +7,7 @@ locals {
   partition           = data.aws_partition.current.partition
   dns_suffix          = data.aws_partition.current.dns_suffix
   region              = data.aws_region.current.name
-  role_name_condition = var.role_name != null ? var.role_name : "${var.role_name_prefix}*"
+  role_name_condition = try(coalesce(var.role_name, "${var.role_name_prefix}*"), null)
 }
 
 data "aws_iam_policy_document" "this" {
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -1,4 +1,42 @@
 ################################################################################
+# API Gateway Controller Policy
+################################################################################
+
+data "aws_iam_policy_document" "api_gateway_controller" {
+  count = var.create_role && var.attach_api_gateway_controller_policy ? 1 : 0
+
+  # https://github.com/aws/aws-application-networking-k8s/blob/main/examples/recommended-inline-policy.json
+  statement {
+    actions = [
+      "vpc-lattice:*",
+      "iam:CreateServiceLinkedRole",
+      "ec2:DescribeVpcs",
+      "ec2:DescribeSubnets",
+    ]
+    resources = ["*"]
+  }
+}
+
+
+resource "aws_iam_policy" "api_gateway_controller" {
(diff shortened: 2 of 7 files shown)


bryantbiggs:
pending name clarification

modules/iam-role-for-service-accounts-eks/policies.tf
```diff
@@ -1,3 +1,41 @@
+################################################################################
+# API Gateway Controller Policy
+################################################################################
+
+data "aws_iam_policy_document" "api_gateway_controller" {
+  count = var.create_role && var.attach_api_gateway_controller_policy ? 1 : 0
+
+  # https://github.com/aws/aws-application-networking-k8s/blob/main/examples/recommended-inline-policy.json
```
antonbabenko, line 8: In the comment, does it make sense to point to a tag/release instead of a moving `main` branch?

The rest LGTM.

bryantbiggs: good point - corrected in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/378/commits/5ca15c6b85f6802b384d6ab9687d136c11ac7a5c

modules/iam-role-for-service-accounts-eks/main.tf
```diff
@@ -7,7 +7,7 @@ locals {
   partition           = data.aws_partition.current.partition
   dns_suffix          = data.aws_partition.current.dns_suffix
   region              = data.aws_region.current.name
-  role_name_condition = var.role_name != null ? var.role_name : "${var.role_name_prefix}*"
+  role_name_condition = try(coalesce(var.role_name, "${var.role_name_prefix}*"), null)
```
antonbabenko, line 10: It will never be `null`, because there is always at least `*`, so `try()` is obsolete.

bryantbiggs: I think you are thinking of the `karpenter_controller_cluster_id`/`karpenter_controller_cluster_name` which defaults to `"*"` - `role_name` and `role_name_prefix` both default to `null` which will cause `coalesce()` to fail if neither are provided a value

bryantbiggs: oh, or perhaps you are referring to the asterisk on the prefix variant - this too fails when `role_name_prefix` is `null`, stating that a `null` value cannot be supplied to a templated value

antonbabenko: Right, then it is well-written. :)

antonbabenko (approved):

antonbabenko:
This PR is included in version 5.20.0 :tada:

PR #358, opened 2023-03-27
feat: Add `elasticloadbalancing:AddTags` permissions to AWS Load Balancer Controller policy required for version 2.4.7+

michelzanini:
Add elasticloadbalancing:AddTags permissions to AWS Load Balancer Controller policy required for version 2.4.7

## Description
Version 2.4.7 of AWS ALB controller has been released with an updated IAM permission https://github.com/kubernetes-sigs/aws-load-balancer-controller/releases/tag/v2.4.7.

The permission required is elasticloadbalancing:AddTags. See change here https://github.com/kubernetes-sigs/aws-load-balancer-controller/pull/3046/files.

## Motivation and Context
Updates AWS ALB Controller policy to match latest on version 2.4.7.

## Breaking Changes
None.

## How Has This Been Tested?
I have applied the IAM policy to AWS using Terraform, then copied the resulting JSON file from AWS, and did a file side-by-side comparison with the official IAM policy JSON confirming it matches.

Fixes #339

PR #382, opened 2023-05-19
feat: Explicitly grant `elasticloadbalancing:AddTags` permission 

acdha:
As per AWS customer notifications, the `elasticloadbalancing:AddTags` permission will be required soon for anyone calling `CreateLoadBalancer` with tags:

> On June 1, 2023, we will be adding an additional layer of security to ELB ‘Create*' API calls where API callers must have explicit access to add tags in their Identity and Access Management (IAM) policy [1]. Currently, access to attach tags was implicitly granted with access to 'Create*' APIs. … We will be allowing 'Create*' API calls with the current policy to be accepted until August 30, 2023. After this date, the 'Create*' API call will fail and return an error if the the attribute is specified and permission is not granted.

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -795,6 +795,7 @@ data "aws_iam_policy_document" "load_balancer_controller" {
 
   statement {
     actions = [
+      "elasticloadbalancing:AddTags",
       "elasticloadbalancing:CreateLoadBalancer",
       "elasticloadbalancing:CreateTargetGroup",
     ]
@@ -809,6 +810,7 @@ data "aws_iam_policy_document" "load_balancer_controller" {
 
   statement {
     actions = [
+      "elasticloadbalancing:AddTags",
       "elasticloadbalancing:CreateListener",
       "elasticloadbalancing:DeleteListener",
       "elasticloadbalancing:CreateRule",


bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.21.0 :tada:

tomelliff:
I don't think this is the equivalent of [the current upstream policy](https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.5.3/docs/install/iam_policy.json) but the translation to HCL makes it a little hard to follow.

The diff between [v2.4.3](https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.4.3/docs/install/iam_policy.json) and v2.5.3 is this statement block:

```diff
+        {
+            "Effect": "Allow",
+            "Action": [
+                "elasticloadbalancing:AddTags"
+            ],
+            "Resource": [
+                "arn:aws:elasticloadbalancing:*:*:targetgroup/*/*",
+                "arn:aws:elasticloadbalancing:*:*:loadbalancer/net/*/*",
+                "arn:aws:elasticloadbalancing:*:*:loadbalancer/app/*/*"
+            ],
+            "Condition": {
+                "StringEquals": {
+                    "elasticloadbalancing:CreateAction": [
+                        "CreateTargetGroup",
+                        "CreateLoadBalancer"
+                    ]
+                },
+                "Null": {
+                    "aws:RequestTag/elbv2.k8s.aws/cluster": "false"
+                }
+            }
+        },
```

which appears to already be in the existing policy having been added in https://github.com/terraform-aws-modules/terraform-aws-iam/commit/e1403c15285137147db2698465e794d17059c848 in https://github.com/terraform-aws-modules/terraform-aws-iam/pull/358.

Would it make more sense to instead use the raw JSON from upstream here to make it much easier to maintain this and show the direct link to upstream?

Issue #383, opened 2023-05-23
try expression makes condition statements fail if role_name_prefix var is not set 


iomarcovalente:
## Description

Latest code release makes condition statements [here](https://github.com/terraform-aws-modules/terraform-aws-iam/blame/e46ff1f49aef00a3c2f95a719560ec86f83c3f5b/modules/iam-role-for-service-accounts-eks/main.tf#L32) fail if `role_name_prefix` var is not set.

Code reference:
https://github.com/terraform-aws-modules/terraform-aws-iam/blame/e46ff1f49aef00a3c2f95a719560ec86f83c3f5b/modules/iam-role-for-service-accounts-eks/main.tf#L10

If your request is for a new feature, please use the `Feature request` template.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## ⚠️ Note

Before you submit an issue, please perform the following first:

1. Remove the local `.terraform` directory (! ONLY if state is stored remotely, which hopefully you are following that best practice!): `rm -rf .terraform/`
2. Re-initialize the project root to pull down modules: `terraform init`
3. Re-attempt your terraform plan or apply and check if the issue still persists

## Versions

- Module version [Required]: 5.20

- Terraform version:
Terraform `v1.2.9`
also tested with 
TF version `v1.4.6`
- Provider version(s):
```
Terraform v1.2.9
on darwin_arm64
+ provider registry.terraform.io/gavinbunney/kubectl v1.14.0
+ provider registry.terraform.io/hashicorp/aws v4.67.0
+ provider registry.terraform.io/hashicorp/cloudinit v2.3.2
+ provider registry.terraform.io/hashicorp/helm v2.9.0
+ provider registry.terraform.io/hashicorp/kubernetes v2.20.0
+ provider registry.terraform.io/hashicorp/local v2.4.0
+ provider registry.terraform.io/hashicorp/null v3.2.1
+ provider registry.terraform.io/hashicorp/tls v3.4.0
+ provider registry.terraform.io/hashicorp/vault v3.15.2
+ provider registry.terraform.io/terraform-aws-modules/http v2.4.1
```

## Reproduction Code [Required]

`locals.tf`
```
locals {
  role_name_prefix = null
}
```
in terraform console, run:
```
try(coalesce("hello", "${local.role_name_prefix}*"), null)
```
output is always `null`

## Expected behavior

the expression to return a non `null` value

## Actual behavior

it does return `null`

## Additional context

[...]

PR #389, opened 2023-06-04
:wrench: Fix local role_name_condition to avoid null result

mstiri:
## Description

When using the `iam-role-for-service-accounts-eks` module with the `allow_self_assume_role` set to `true` a dynamic statement is generated for the policy. This policy uses a local role_name_condition computed using the coalesce function as follows:
`coalesce(var.role_name, "${var.role_name_prefix}*")`
The issue comes from the fact that the `var.role_name_prefix` is `null` by default, and cannot be included in a string template. The error generated is:
`The expression result is null. Cannot include a null value in a string template.`

## Motivation and Context
This change solves the issue generated when the `allow_self_assume_role` set to `true`. It is a fix to handle the different cases properly
The PR also fixes this open issue: #383 

## Breaking Changes
The PR does not introduce any breaking changes. 

## How Has This Been Tested?

Changes have been tested by creating a new example taking the `module.cert_manager_irsa_role` as a base and setting `allow_self_assume_role` to `true`. The new example is included in the PR to make sure this case is always properly handled, it is under `module.cert_manager_irsa_role_self_assume`
The original example has been tested with the new changes to make sure nothing is broken. 

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/main.tf
+++ b/modules/iam-role-for-service-accounts-eks/main.tf
@@ -7,7 +7,7 @@ locals {
   partition           = data.aws_partition.current.partition
   dns_suffix          = data.aws_partition.current.dns_suffix
   region              = data.aws_region.current.name
-  role_name_condition = try(coalesce(var.role_name, "${var.role_name_prefix}*"), null)
+  role_name_condition = var.role_name != null ? var.role_name : "${var.role_name_prefix}*"
 }
 
 data "aws_iam_policy_document" "this" {
--- a/examples/iam-role-for-service-accounts-eks/main.tf
+++ b/examples/iam-role-for-service-accounts-eks/main.tf
@@ -85,6 +85,24 @@ module "cert_manager_irsa_role" {
   tags = local.tags
 }
 
+module "cert_manager_irsa_role_self_assume" {
+  source = "../../modules/iam-role-for-service-accounts-eks"
+
+  role_name                     = "cert-manager"
+  attach_cert_manager_policy    = true
+  cert_manager_hosted_zone_arns = ["arn:aws:route53:::hostedzone/IClearlyMadeThisUp"]
+  allow_self_assume_role        = true
+
+  oidc_providers = {
+    ex = {
+      provider_arn               = module.eks.oidc_provider_arn
+      namespace_service_accounts = ["kube-system:cert-manager"]
+    }
+  }
+
+  tags = local.tags
+}
+
 module "cluster_autoscaler_irsa_role" {
   source = "../../modules/iam-role-for-service-accounts-eks"
 


modules/iam-role-for-service-accounts-eks/main.tf
```diff
@@ -7,7 +7,7 @@ locals {
   partition           = data.aws_partition.current.partition
   dns_suffix          = data.aws_partition.current.dns_suffix
   region              = data.aws_region.current.name
-  role_name_condition = try(coalesce(var.role_name, "${var.role_name_prefix}*"), null)
+  role_name_condition = var.role_name != null ? var.role_name : "${var.role_name_prefix}*"
```
bryantbiggs, line 10: this won't solve the issue - to solve the issue we just need to change the default value of `role_name_prefix` to `""`

mstiri: Setting the `role_name_prefix` to `""` does not solve the issue. It will create another issue because you will have `role_name` set to a value and `role_name_prefix` set to `""`, and providing both of them to `aws_iam_role` generates a conflict error:
`"name_prefix": conflicts with name`

ViktorTverdun: I think `aws_iam_role` can be created without either `name` or `name_prefix` which means both `role_name` and `role_name_prefix` can be kept  with default values `null`. In this case, "else" part will be executed and `"${var.role_name_prefix}*"` produces same error `Cannot include a null value in a string template.`
What do you think about changing exactly the problematic place? I mean, something like
```
role_name_condition = var.role_name != null ? var.role_name : "${var.role_name_prefix == null ? "" : var.role_name_prefix}*"
```
This way if role is created without name or prefix, `role_name_condition` is `*` but never null

mstiri: But we want to allow only this role to assume itself. 
```
condition {
        test     = "ArnLike"
        variable = "aws:PrincipalArn"
        values   = ["arn:${local.partition}:iam::${local.account_id}:role${var.role_path}${local.role_name_condition}"]
}
```
With `role_name_condition` set to `*,` the policy is too permissive.
The suggested solution solves the issue. Why is it problematic?

ViktorTverdun: Hm, you're right about too permissive. By problematic I meant the place that causes the error

mstiri: Yeah, I see. But what is wrong with the proposed solution?

ViktorTverdun: If module is initiated without specifying `role_name` and `role_name_prefix` they are `null` by default. So `var.role_name != null` is false and "else" statement, which is `"${var.role_name_prefix}*"` being executed. This produces error "Cannot include a null value in a string template."  even if `allow_self_assume_role` is false.

ViktorTverdun: In general, this solution is the same is it was in 5.19.0, so should be ok. 
Was I set as review on purpose or it's automatic?

mstiri: Yes, this solution is the same as it was in previous versions. And it is the same as how it is handled in other modules, such as: [iam-assumable-role-with-oidc/main.tf](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-assumable-role-with-oidc/main.tf#L10)

bdellegrazie:
Is this going to get merged and released?

bryantbiggs (approved):

PR #398, opened 2023-06-28
fix: Hardcode known intermediary thumbprints in github oidc provider

therealdwright:
## Description
Following GitHub's announcement that can be found on their [blog](https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/) there are now multiple thumbprints that are required to integrate AWS with GitHub actions.

We have been running with this change for some time now and wanted to submit upstream in case others are having issues with their GitHub actions workflows assuming AWS roles.

## Motivation and Context
Solves errors where GitHub actions workflows are unable to assume an IAM role due to an invalid thumbprint.

## Breaking Changes
N/A - this change is additive and won't interfere with the current implementation.

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s) (N/A)

As previously mentioned, we have been testing this change for a while. We vendored this change into our own Terraform workspace and have been running without issue for a while now. During the last couple of weeks, GitHub has changed thumbprints twice and our change detected when this occurred and showed up in a Terraform plan.

### DIFF
--- a/modules/iam-github-oidc-provider/README.md
+++ b/modules/iam-github-oidc-provider/README.md
@@ -30,7 +30,6 @@ module "iam_github_oidc_provider" {
 | Name | Version |
 |------|---------|
 | <a name="provider_aws"></a> [aws](#provider\_aws) | >= 4.0 |
-| <a name="provider_tls"></a> [tls](#provider\_tls) | >= 3.0 |
 
 ## Modules
 
@@ -42,7 +41,6 @@ No modules.
 |------|------|
 | [aws_iam_openid_connect_provider.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_openid_connect_provider) | resource |
 | [aws_partition.current](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/partition) | data source |
-| [tls_certificate.this](https://registry.terraform.io/providers/hashicorp/tls/latest/docs/data-sources/certificate) | data source |
 
 ## Inputs
 
@@ -50,6 +48,7 @@ No modules.
 |------|-------------|------|---------|:--------:|
 | <a name="input_client_id_list"></a> [client\_id\_list](#input\_client\_id\_list) | List of client IDs (also known as audiences) for the IAM OIDC provider. Defaults to STS service if not values are provided | `list(string)` | `[]` | no |
 | <a name="input_create"></a> [create](#input\_create) | Controls if resources should be created (affects all resources) | `bool` | `true` | no |
+| <a name="input_github_intermediary_thumbprints"></a> [github\_intermediary\_thumbprints](#input\_github\_intermediary\_thumbprints) | GitHub actions SSL intermeidary thumbprints | `set(string)` | <pre>[<br>  "6938fd4d98bab03faadb97b34396831e3780aea1",<br>  "1c58a3a8518e8759bf075b76b750d4f2df264fcd"<br>]</pre> | no |
 | <a name="input_tags"></a> [tags](#input\_tags) | A map of tags to add to the resources created | `map(any)` | `{}` | no |
 | <a name="input_url"></a> [url](#input\_url) | The URL of the identity provider. Corresponds to the iss claim | `string` | `"https://token.actions.githubusercontent.com"` | no |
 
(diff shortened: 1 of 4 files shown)


antonbabenko (changes requested):
I don't see how this differs from what we already have in this module.

For me, these two data sources produce the same results:
```hcl
data "tls_certificate" "new" {
  url = "https://token.actions.githubusercontent.com/.well-known/openid-configuration"
}
```

```hcl
data "tls_certificate" "current" {
  url = "https://token.actions.githubusercontent.com"
}
```

```
sha1_fingerprint = [
  "6938fd4d98bab03faadb97b34396831e3780aea1",
  "f879abce0008e4eb126e0097e46620f5aaae26ad",
]
```

`1c58a3a8518e8759bf075b76b750d4f2df264fcd` is missing for me.

As discussed in [this comment](https://github.com/aws-actions/configure-aws-credentials/issues/357#issuecomment-1610172322), I think we should do a call to `distinct(concat(...))` to get a list of at least 3 fingerprints (assuming GH will roll the updates in the future and we don't need to update it again).

Will you be able to update the code accordingly?

therealdwright:
> Every second request to the data source changes

Thank you for bringing this to my attention I thought I was going crazy and getting a bit better stability with the endpoint I provided, but it was wrong. I think I agree that at least in the interim until these fingerprints are stable that hardcoding is the right approach. I will update the PR to use the fingerprints provided by GitHub and leave it to the maintainers to decide if they want a TODO or to merge at all.

bryantbiggs:
Here we can simply wrap it in a `sort()` to make the diff the same - we should NOT hardcode these

arianvp:
> Here we can simply wrap it in a sort() to make the diff the same - we should NOT hardcode these

This is simply not possible.  Github is returning a different intermediate cert on each request. How do you want to not hardcode this?

bryantbiggs:
> This happens for customers who are pinned to a single intermediary thumbprint from the Certificate Authority (CA) of the Actions SSL certificate.

Pinning something that is known to change, and designed to change, is not a viable solution

therealdwright:
> Pinning something that is known to change, and designed to change, is not a viable solution

Let me know if you would like this PR closed. We have pinned our own custom implementation as this provider is not usable in its current form.

The commit message states it's an interim solution.

bryantbiggs:
why is it not usable in its current form? its returning the wrong thumbprints? its returning the incorrect thumbprints?

therealdwright:
> why is it not usable in its current form? its returning the wrong thumbprints? its returning the incorrect thumbprints?

The provider returns different thumbprints on every CI run leading to detected changes on every single Terraform plan. We run CI automation that plans our infrastructure every PR to our CI environment and it's not viable for us to have unrelated changes in every single PR.

therealdwright:
If you would like me to go with a conditional approach where pinned thumbprints are an optional variable I will do this. However, I'm not going to invest any more time until I get some consensus from maintainers on whether or not this is something they want to solve.

bryantbiggs:
while spurious diffs are annoying, I would prefer to have that rather than having a hardcoded value that will change and be incorrect at some point in the future. I think the proper solution here is for GitHub to provide the guidance or changes to alleviate this issue. As Anton mentioned above, there is a whole discussion happening here https://github.com/aws-actions/configure-aws-credentials/issues/357 with no resolution currently

flokli:
@bryantbiggs you misunderstood. Github actions HTTPS endpoint uses either one certificate chain or another certificate chain for different HTTPS connections, it's pure luck which one you end up with.

A single terraform run might end up discovering one fingerprint, while another one will discover the other one. There is simply no way to properly discover this from Terraform in a reliable fashion, see https://github.com/aws-actions/configure-aws-credentials/issues/357#issuecomment-1612005962 for a detailed answer.

The proper fix would be AWS "Internal P82323250" issue, but in the meantime, pinning certs as documented by AWS in https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/ is the officially documented way, so that's what this terraform module should do too.

therealdwright:
> A single terraform run might end up discovering one fingerprint, while another one will discover the other one

I think the maintainers understand the issue they just don't agree with pinning the images or even providing the option to do so. It's, for this reason, we decided to abandon using the module.

PR #403, opened 2023-07-02
feat: Github OIDC add extra thumbprints as needed

### LINKED ISSUE #400: iam-github-oidc-provider should allow for passing additional thumbprint to remain compatible with recent changes to Github Actions OIDC provider
[Github has recently released a blog post](https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/) detailing a potential issue and workaround for certificates related to their Identity Provider for OIDC and their Actions provider for AWS authentication.  Basically, you have a chance of receiving one of two possible certificates (in a round robin fashion) when trying to get OIDC credentials from a GHA pipeline and may experience intermittent failures in CI pipelines as a result.  

How this pertains to this repo is that when using the `iam-github-oidc-provider` module, it _**only**_ uses the thumbprints for the cert that it picks up on a given run (either/or) and doesn't allow for supplying additional thumbprints to the module.  I have the behavior display as it changing back and forth on repeat runs like so (both the value to be added and the value to be removed here are valid thumbprints):

```
  # module.aws_account.module.iam_github_oidc_provider.aws_iam_openid_connect_provider.this[0] will be updated in-place
  ~ resource "aws_iam_openid_connect_provider" "this" {
        id              = "arn:aws:iam::003929400660:oidc-provider/token.actions.githubusercontent.com"
        tags            = {
            "Environment" = "Global"
        }
      ~ thumbprint_list = [
          - "6938fd4d98bab03faadb97b34396831e3780aea1",
          + "1c58a3a8518e8759bf075b76b750d4f2df264fcd",
            "f879abce0008e4eb126e0097e46620f5aaae26ad",
        ]
        # (4 unchanged attributes hidden)
    }
```

There should be a facility for passing additional thumbprints to the module via an optional var and then those thumbprints can be added to the `thumbprint_list` value of the resource.  It would be a good idea to convert the list to a set and back to a list in the process to eliminate duplicates.

The only alternative I see at present is to not use the module or manually change the thumbprint list as needed in AWS, which pretty much defeats the purpose of using Terraform.

## Additional context
I'm fully aware that the root of the issue is caused by Github behavior but the module should be compatible and it should be a relatively simple feature/fix.

magreenbaum:
## Description
Add additional thumbprint list as needed. 

## Motivation and Context
- Resolves #400

## Breaking Changes
No.

## How Has This Been Tested?

### DIFF
--- a/examples/iam-github-oidc/main.tf
+++ b/examples/iam-github-oidc/main.tf
@@ -21,12 +21,6 @@ locals {
 module "iam_github_oidc_provider" {
   source = "../../modules/iam-github-oidc-provider"
 
-  # https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/
-  additional_thumbprint_list = [
-    "6938fd4d98bab03faadb97b34396831e3780aea1",
-    "1c58a3a8518e8759bf075b76b750d4f2df264fcd"
-  ]
-
   tags = local.tags
 }
 
--- a/examples/iam-github-oidc/main.tf
+++ b/examples/iam-github-oidc/main.tf
@@ -21,6 +21,12 @@ locals {
 module "iam_github_oidc_provider" {
   source = "../../modules/iam-github-oidc-provider"
 
+  # https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/
+  additional_thumbprint_list = [
+    "6938fd4d98bab03faadb97b34396831e3780aea1",
+    "1c58a3a8518e8759bf075b76b750d4f2df264fcd"
+  ]
+
   tags = local.tags
 }
 
--- a/modules/iam-github-oidc-provider/variables.tf
+++ b/modules/iam-github-oidc-provider/variables.tf
@@ -21,3 +21,9 @@ variable "url" {
   type        = string
   default     = "https://token.actions.githubusercontent.com"
 }
+
+variable "additional_thumbprint_list" {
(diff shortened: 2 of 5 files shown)


examples/iam-github-oidc/main.tf
```diff
@@ -21,6 +21,12 @@ locals {
 module "iam_github_oidc_provider" {
   source = "../../modules/iam-github-oidc-provider"
 
+  # https://github.blog/changelog/2023-06-27-github-actions-update-on-oidc-integration-with-aws/
+  additional_thumbprint_list = [
```
antonbabenko, line 25: Instead of putting these values in the examples, we would rather have them as defaults, so users should just upgrade the version of the module and use it right away.

/cc @bryantbiggs WDYT?

bryantbiggs: ya we can try that. I've checked internally as well and there isn't a clear path to resolution so lets try and see what happens

modules/iam-github-oidc-provider/variables.tf
```diff
@@ -21,3 +21,9 @@ variable "url" {
   type        = string
   default     = "https://token.actions.githubusercontent.com"
 }
+
+variable "additional_thumbprint_list" {
```
bryantbiggs, line 25: lets just call this 
```suggestion
variable "additional_thumbprints" {
```

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.26.0 :tada:

PR #411, opened 2023-08-07
feat(iam-eks-role): Add the ability to provide additional principal ARNs which could assume the role

resurtm:
## Description

Add the ability to provide additional principal ARNs which could assume the role.

## Motivation and Context

Please check the additions to the README.md file. I've tested this one with my private project with the real AWS setup and Kubernetes cluster, and it works correctly. I tried to come up with some solution without modifying the upstream module and do all the changes in my private Git repository, but I don't really have any ideas how could I override the `assume_role_policy` value (of the `aws_iam_role` resource) in a nice way without copy'n'pasting half of this module. If you have better ideas, please let me know. Thank you!

## Breaking Changes

There are no breaking changes. The proposed code and changes should be backward compatible with any kind of existing Terraform code that uses this module.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-eks-role/README.md
+++ b/modules/iam-eks-role/README.md
@@ -74,6 +74,29 @@ module "iam_eks_role" {
 }
 ```
 
+## Provide additional principal ARNs which could assume the managed role
+
+In some cases you might need to provide additional custom principal ARNs which should be able to assume the role.
+This kind of principal ARNs must be provided only in the `assume_role_policy` of the `aws_iam_role` resource.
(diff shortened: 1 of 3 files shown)


modules/iam-eks-role/README.md
```diff
@@ -74,6 +74,29 @@ module "iam_eks_role" {
 }
 ```
 
+## Provide additional principal ARNs which could assume the managed role
+
+In some cases you might need to provide additional custom principal ARNs which should be able to assume the role.
```
bryantbiggs, line 79: What cases?

PR #412, opened 2023-08-16
chore: Update VPC CNI policy to use scope down permissions

matheusmazzoni:
## Description
This implementation causes the VPC CNI policy to be updated to give fewer permissions and mitigate the possibilities of attacks using the policy.

## Motivation and Context

The motivation is in this documentation: [ABUSING AMAZON VPC CNI PLUGIN FOR KUBERNETES](https://www.elttam.com/blog/amazon-vpc-cni/#content). Due to the wild card ["Resource": "*"](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts-eks/policies.tf#L1367) on the VPC CNI policy and lack of any restricting conditions, the policy allows manipulating the EC2 networking of all EC2 instances, not just those in the node groups for the EKS cluster that was created with eksctl.
Therefore, the policy was modified so that it is only possible to manage EKS nodes within the EKS VPC.

## How Has This Been Tested?


bryantbiggs:
thank you for the PR, but I am hesitant to deviate from the policy provided by the VPC CNI project. we have seen challenges when we have done this in the past (Karpenter, AWS Load Balancer Controller) which leads to a thrashing of changes

matheusmazzoni:
> thank you for the PR, but I am hesitant to deviate from the policy provided by the VPC CNI project. we have seen challenges when we have done this in the past (Karpenter, AWS Load Balancer Controller) which leads to a thrashing of changes

I completely understand @bryantbiggs! Here is an example from the AWS documentation within the aws-vpc-cni-k8s repository that talks about using scope down policy on VPC CNI K8s implementation: [Scope Down IAM policy VPC CNI](https://github.com/aws/amazon-vpc-cni-k8s/blob/master/docs/iam-policy.md#scope-down-iam-policy-per-eks-cluster).

If you think it's feasible, I can leave the old implementation and add a new logic that makes it possible to use scope down optionally, for example:
```
vpc_cni_enable_ipv4 = true
vpc_cni_use_scope_down_policy = true
```
We currently have this use case, as I don't see why we don't have an additional security option. What do you think 😸?

bryantbiggs:
yes, we can support both with some dynamic conditionals. Lets use the variable `vpc_cni_enable_scoped_policy`. Do you know if this supports both IPv4 and IPv6 or just IPv4?

matheusmazzoni:
@bryantbiggs I'll add the variable `vpc_cni_enable_scoped_policy` for implement the low scope policy. Regarding your question about IPV6, I have no knowledge about it, maybe it works a little differently, but I'll have to test it to give you feedback. Initially we can leave it only for IPv4 and I can open another PR after testing with IPv6. What you think?

matheusmazzoni:
@bryantbiggs Would you give the permission to do firstly just a ipv4 implementation?

Issue #408, opened 2023-07-26
 'iam-user' new feature 'policy_arns' not working with 'iam-policy' ARN ouput #246 


BetterToAutomateTheWorld:
Hi all,

First of all, thank you for these incredible TF modules ! :)

## Description

"policy_arns" parameters in "terraform-aws-modules/iam/aws//modules/iam-user" seems not working as expected when using 'iam-policy' module .arn output 
I don't know if it's due to the "policy_arns" from iam-user or to the "arn" output from iam-policy not formatted the right way, or the usage of both together that create issues

- [x] ✋ I have searched the open/closed issues and my issue is not listed.
--> New iam-user module feature from 1.58.0, no issue found yet !

## Versions

- Module version [Required]:
1.58.0
- Terraform version:
Terraform v1.5.3
- Provider version(s):
provider registry.terraform.io/datadog/datadog v3.28.0
provider registry.terraform.io/hashicorp/aws v4.67.0
provider registry.terraform.io/hashicorp/external v2.3.1
provider registry.terraform.io/hashicorp/local v2.4.0
provider registry.terraform.io/hashicorp/null v3.2.

## Reproduction Code [Required]

```
module "iam-policy-bucket-xxx-local" {
  source  = "terraform-aws-modules/iam/aws//modules/iam-policy"
  version = "5.28.0"

  name        = "xxx-local-policy"
  path        = "/"
  description = "Policy to fully access the specified S3 bucket(s)"

  policy = <<EOF
{
  "Version": "2012-10-17",
  "Statement": [
		{
			"Sid": "S3All",
			"Effect": "Allow",
			"Action": [
        "s3:PutObject",
        "s3:GetObject",
        "s3:ListBucket",
        "s3:GetBucketVersioning",
        "s3:AbortMultipartUpload",
        "s3:ListMultipartUploadParts",
        "s3:ListBucketMultipartUploads",
        "s3:DeleteObject",
        "s3:GetBucketLocation",
        "s3:GetObjectVersion"
			],
			"Resource": [
        "${module.bucket-xxx-local.s3_bucket_arn}",
        "${module.bucket-xxx-local.s3_bucket_arn}/*"
			]
		}
  ]
}
EOF
}

module "xxx-local-user" {
  source  = "terraform-aws-modules/iam/aws//modules/iam-user"
  version = "5.28.0"

  name = "xxx-local"

  force_destroy           = true
  password_reset_required = false

[...]

PR #421, opened 2023-09-04
feat(iam-user): Fixed direct policy attachment in iam-user module

garylau:
## Description
When creating the `iam-user` module, it throws error as `for_each` is expecting a map or a set of string however the module defines the `policy_arns` variable as a `list(string)`.

## Motivation and Context
Fixes https://github.com/terraform-aws-modules/terraform-aws-iam/issues/408

There are two options to fix this issue. Option 1 is to cast the `policy_arns` into a map of string in runtime (like how it is done in https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-eks-role/main.tf#L78). Or option 2 is to replace `for_each` with `count`. For simplicity reason, I chose option 2.

## Breaking Changes
None

## How Has This Been Tested?
example remains the same as there is no change to the caller inputs

### DIFF
--- a/modules/iam-user/main.tf
+++ b/modules/iam-user/main.tf
@@ -47,8 +47,8 @@ resource "aws_iam_user_ssh_key" "this" {
 }
 
 resource "aws_iam_user_policy_attachment" "this" {
-  for_each = var.create_user ? toset(var.policy_arns) : []
+  count = var.create_user ? length(var.policy_arns) : 0
 
   user       = aws_iam_user.this[0].name
-  policy_arn = each.value
+  policy_arn = element(var.policy_arns, count.index)
 }


garylau:
@antonbabenko hello, any chance you may take a look? thank you!

igor-nikiforov:
@antonbabenko @bryantbiggs could you please take a look this PR?

antonbabenko (changes requested):
Sorry for the delay with the review. Let's stick to `for_each` and make it changed as here - https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-eks-role/main.tf#L78

diosif-cr:
@antonbabenko I've created a PR with your proposal, as I've stumbled upon this issue today.

https://github.com/terraform-aws-modules/terraform-aws-iam/pull/428

Issue #422, opened 2023-09-13
Add CIDR limitation to OIDC IAM Role


alecsiemerink:
## Is your request related to a new offering from AWS?
Is this functionality available in the [AWS provider for Terraform](https://registry.terraform.io/providers/hashicorp/aws/latest/docs)? See [CHANGELOG.md](https://github.com/hashicorp/terraform-provider-aws/blob/main/CHANGELOG.md), too.
  - Yes ✅: At least > V3
## Is your request related to a problem? Please describe.

We want to secure our Terraform deployment pipelines which already use OIDC on a self-hosted GitHub Actions runner. We want to be able to limit the CIDR from which this role can be used. 

## Describe the solution you'd like.

In the module declaration, I want to be able to pass a CIDR list like this:
```
module "iam_github_oidc_role" {
  source    = "terraform-aws-modules/iam/aws//modules/iam-github-oidc-role"

  # This should be updated to suit your organization, repository, references/branches, etc.
  subjects = ["terraform-aws-modules/terraform-aws-iam:*"]

  policies = {
    S3ReadOnly = "arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess"
  }
  allowed_cidr = ['x.x.x.x/x", "x.x.x.x/x"]


  tags = {
    Environment = "test"
  }
}
```

## Describe alternatives you've considered.

Doing this through the console:
![image](https://github.com/terraform-aws-modules/terraform-aws-iam/assets/41921577/07b71e55-c99a-4432-8401-186b8539997c)

PR #423, opened 2023-09-13
feat(iam-github-oidc-role): Add condition for allowlisting CIDR.

alecsiemerink:
## Description

This allows you to pass a list of CIDRs from where the role can be used. 
A non-required variable is introduced, that when it is set it will add a condition to the `aws_iam_policy_document` with the CIDR. 

## Motivation and Context

https://github.com/terraform-aws-modules/terraform-aws-iam/issues/422
This enables more secure use of OIDC by allowing only a specified CIDR to use it. For example, for using self-hosted GitHub Actions runners and allowlisting them. 
## Breaking Changes

Not that I know of.
## How Has This Been Tested?


alecsiemerink:
No activity here?

PR #440, opened 2023-11-27
fix(iam-group-with-policies): Related resources shouldn't be created when group creation is disabled

pawelpesz:
## Description
Related resources shouldn't be created when group creation is disabled

## Motivation and Context
In some scenarios the `iam-group-with-policies` submodule attempts to create `aws_iam_group_membership`, `aws_iam_policy` and `aws_iam_group_policy_attachment` resources even though the `create_group` variable is set to `false`. Workarounds do exist, but ultimately this behaviour is undesired.

## Breaking Changes
None.

## How Has This Been Tested?

### DIFF
--- a/examples/iam-group-with-policies/main.tf
+++ b/examples/iam-group-with-policies/main.tf
@@ -65,6 +65,35 @@ module "iam_group_with_custom_policies" {
   ]
 }
 
+#####################################################################################
+# IAM group to test the `create_group = false` option
+#####################################################################################
+module "iam_group_with_custom_policies_disabled" {
+  source = "../../modules/iam-group-with-policies"
+
+  create_group = false
+
+  name = "custom-disabled"
+  path = "/custom/"
+
+  group_users = [
+    module.iam_user1.iam_user_name,
+    module.iam_user2.iam_user_name,
+  ]
+
+  custom_group_policy_arns = [
+    "arn:aws:iam::aws:policy/AmazonCognitoReadOnly",
+    "arn:aws:iam::aws:policy/AlexaForBusinessFullAccess",
+  ]
+
+  custom_group_policies = [
+    {
+      name   = "AllowS3Listing"
+      policy = data.aws_iam_policy_document.sample.json
+    },
+  ]
+}
+
 ######################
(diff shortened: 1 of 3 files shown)


pawelpesz:
Not stale, just updated.

pawelpesz:
Not stale.

pawelpesz:
Hello @bryantbiggs, sorry to bother you, but could you take a quick look at this PR? It's nothing controversial, just a simple fix for something that got overlooked. Thank you!

examples/iam-group-with-policies/main.tf
```diff
@@ -65,6 +65,35 @@ module "iam_group_with_custom_policies" {
...
 
+#####################################################################################
+# IAM group to test the `create_group = false` option
+#####################################################################################
+module "iam_group_with_custom_policies_disabled" {
+  source = "../../modules/iam-group-with-policies"
+
+  create_group = false
+
+  name = "custom-disabled"
```
antonbabenko, line 76: When `create_group` is set to `false,` then it expects that the IAM group with the name specified in `name` has already been created. Also, this module adds provided properties (such as `group_users`, `custom_group_policy_arns`, ...) to it.

antonbabenko (commented):
This is not an issue, but an expected behavior of the module.

examples/iam-group-with-policies/main.tf
pawelpesz: @antonbabenko with due respect, but I don't see the same the behaviour being followed in the other modules, `iam-user` for example.

antonbabenko: If we had called this module `iam-group,` then the scope of resources should be - `all` or `nothing` (like in `iam-user`), but since it is `iam-group-with-policies`, it expects IAM group in one way or another.

pawelpesz: I found this behaviour pretty confusing @antonbabenko, but fair enough, thank you for the explanation 🙏

antonbabenko: I agree it is confusing.

PR #452, opened 2024-01-16
docs: Remove the invalid `provider` argument from the `oidc_providers` variable description

AndrewCharlesHay:
## Description
The `provider` attribute is never accessed and thus shouldn't be listed. It also changes the type of `oidc_providers` from `any` to a `map(any)` because it is a map.

## Motivation and Context
It clutters the requirements for the variable and thus should be removed

## Breaking Changes
If someone used to pass something into `oidc_providers` it will now throw an error at an earlier stage of validation

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/main.tf
+++ b/modules/iam-role-for-service-accounts-eks/main.tf
@@ -39,7 +39,6 @@ data "aws_iam_policy_document" "this" {
     for_each = var.oidc_providers
 
     content {
-      sid     = "OIDCRoleAssumption${index(var.oidc_providers, each.value) + 1}"
       effect  = "Allow"
       actions = ["sts:AssumeRoleWithWebIdentity"]
 
--- a/modules/iam-role-for-service-accounts-eks/main.tf
+++ b/modules/iam-role-for-service-accounts-eks/main.tf
@@ -39,6 +39,7 @@ data "aws_iam_policy_document" "this" {
     for_each = var.oidc_providers
 
     content {
+      sid     = "OIDCRoleAssumption${index(var.oidc_providers, each.value) + 1}"
       effect  = "Allow"
       actions = ["sts:AssumeRoleWithWebIdentity"]
 
--- a/modules/iam-role-for-service-accounts-eks/variables.tf
+++ b/modules/iam-role-for-service-accounts-eks/variables.tf
@@ -48,7 +48,7 @@ variable "role_policy_arns" {
 
 variable "oidc_providers" {
   description = "Map of OIDC providers where each provider map should contain the `provider_arn` and `namespace_service_accounts`"
-  type        = map(any)
+  type        = any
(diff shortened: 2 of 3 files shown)


AndrewCharlesHay:
I didn't mean to commit the SID addition to this PR. I can make a new PR if you want.

modules/iam-role-for-service-accounts-eks/main.tf
```diff
@@ -39,6 +39,7 @@ data "aws_iam_policy_document" "this" {
     for_each = var.oidc_providers
 
     content {
+      sid     = "OIDCRoleAssumption${index(var.oidc_providers, each.value) + 1}"
```
bryantbiggs, line 42: please revert this change

modules/iam-role-for-service-accounts-eks/variables.tf
```diff
@@ -47,8 +47,8 @@ variable "role_policy_arns" {
 }
 
 variable "oidc_providers" {
-  description = "Map of OIDC providers where each provider map should contain the `provider`, `provider_arn`, and `namespace_service_accounts`"
-  type        = any
```
bryantbiggs, line 51: this is intentional - please revert this change

bryantbiggs (changes requested):

bryantbiggs (approved):
thank you

AndrewCharlesHay:
@bryantbiggs Thank you for the merge! I understand if you want them in a separate PR. But what is wrong with changing the type to `map(any)` instead of just `any`?

bryantbiggs:
when you go with a type, other than `any`, you are forcing users into using the same shape on every input. so for example, if you had something like the following that was acceptable for use by the underlying resource, if you used `any` it will work, if you used `map(any)`, it will complain that the objects are of different types (one has `colors` and the other does not):

```hcl
  foo = {
    one = {
      hot = true
      cups = 2
      colors = ["orange", "blue"]
    }
    two = {
      hot = false
      cups = 1
    }
  }
```

So we typically just default to `any` for complex types so that users have the most flexibility and can provide only the values they wish

AndrewCharlesHay:
Ohh yeah I've run into that before. So we would want to ensure that every object has just `provider_arn` and `namespace_service_accounts` right? I guess the issue would be when the key, in your example `one` and `two` would be different

AndrewCharlesHay:
Both attributes are referenced in each [iteration](https://github.com/AndrewCharlesHay/terraform-aws-iam/blob/e63dff8de9f6aee7e9ebfbe8c676fa980f5233c0/modules/iam-role-for-service-accounts-eks/main.tf#L52) so we need both and wouldn't want anything else right? I guess it would somewhat unnecessarily make things not backwards compatible.

antonbabenko:
This PR is included in version 5.33.1 :tada:

PR #453, opened 2024-01-16
feat: Add default description for role

AndrewCharlesHay:
## Description
Creates a default role description for the oidc provider

## Motivation and Context
Creates more clarity and traceability for IAM roles by default

## Breaking Changes
None

## How Has This Been Tested?


bryantbiggs:
thank you but this is not valid - we can't have a default description that is only applicable to one of the scenarios supported by this sub-module

AndrewCharlesHay:
> thank you but this is not valid - we can't have a default description that is only applicable to one of the scenarios supported by this sub-module

oh whoops you're right. I didn't think about the fact that it's used for other scenarios. Would it be ok if it checked the `attach_external_dns_policy` and other similar variables or no?

bryantbiggs:
what problem are you trying to solve?

PR #287, opened 2022-10-12
fix: Allow TagUser to SelfManagement policy

BastienGenestal:
## Description
Adding the following permissions to IAM users created from the iam-group-with-policies module. 
```
"iam:ListUserTags"
"iam:TagUser"
"iam:UntagUser"
```
For instance, user will now be able to label their own access keys.

## Motivation and Context
https://github.com/terraform-aws-modules/terraform-aws-iam/issues/286

## How Has This Been Tested?
- [ ] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [ ] I have tested and validated these changes using one or more of the provided `examples/*` projects

- [ ] I have executed `pre-commit run -a` on my pull request

PR #461, opened 2024-02-26
feat: Allow users to set and read own access keys description (iam-group-with-policies)

vutny:
## Description
This PR fixes default IAM permissions for user group created by `iam-group-with-policies` module, extending it to allow users set and read own access keys description when creating them via AWS Console.

## Motivation and Context
The issue with missing permissions was already addressed before by PR #287 , but the lines looks like were just lost after some major changes.

## Breaking Changes
None.

## How Has This Been Tested?
By applying the change to existing policies and checking AWS console.


antonbabenko:
This PR is included in version 5.35.0 :tada:

PR #479, opened 2024-04-19
feat: Add support for inline policy creation

fatmcgav:
## Description
This commit adds support for creating an inline policy for an IAM role to the `iam-assumable-role-with-oidc` and `iam-assumable-role` modules.

Also updated examples for these modules.

## Motivation and Context
Allows creation of an IAM role with an inline policy. 
- Resolves #463 
- Resolves #419 
- Resolves #261 

## Breaking Changes
None

## How Has This Been Tested?
- [x] I have updated at least one of the `examples/*` to demonstrate and validate my change(s)
- [x] I have tested and validated these changes using one or more of the provided `examples/*` projects
- [x] I have executed `pre-commit run -a` on my pull request

Issue #426, opened 2023-10-02
Mitigate possible security issue for iam-role-for-service-accounts-eks


marcispauls:
## Is your request related to a problem? Please describe.
It's a hypothetical problem scenario in case of compromised keys (used for oidc authentication).
An attacker could then try to authenticate from anywhere and assume roles by impersonating the service account.
The condition "aws:SourceVpc" should limit the usability of the corresponding IAM role to just the environment it's used (The VPC).

## Describe the solution you'd like.
Provide possibility to configure aws:SourceVpc for iam role created in https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts-eks/main.tf to limit such possibility

The output of trust policy could be something like:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam::ACCOUNT_ID:oidc-provider/OIDC_PROVIDER_URL"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "OIDC_PROVIDER_URL:sub": [
            "system:serviceaccount:SERVICE_ACCOUNT_NAME"
          ],
          "aws:SourceVpc": "VPC_ID"
        }
      }
    }
  ]
}
```

PR #480, opened 2024-04-19
feat: Add support for OIDC policy conditions

fatmcgav:
## Description

This commit adds support to the `iam-assumable-role-with-oidc` module
for adding policy conditions constraints [1]

Also add an example showing use with `aws:RequestTag` session tag match [2].

[1] https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#condition
[2] https://docs.aws.amazon.com/IAM/latest/UserGuide/id_session-tags.html

## Motivation and Context
AWS STS session tags allow for additional control when using the `AssumeRoleWithWebIdentity` authentication call.

## Breaking Changes
None

## How Has This Been Tested?


bryantbiggs:
this looks like you are looking for the https://github.com/terraform-aws-modules/terraform-aws-iam/tree/master/modules/iam-role-for-service-accounts-eks instead - using that sub-module, is this change still warranted? I'm thinking no

fatmcgav:
Hey @bryantbiggs ...

So I think this change is still valid, as I'm using session tags in combination with an external OIDC provider, in this case env0. Docs are here: https://docs.env0.com/docs/oidc-with-aws#custom-claims-with-aws-session-tags-optional

fatmcgav:
@bryantbiggs / @antonbabenko Any chance of a review on this one? 👍

yaroslav-nakonechnikov:
ping

yaroslav-nakonechnikov:
as i understand, it also covers https://github.com/terraform-aws-modules/terraform-aws-iam/issues/426 in different way

fatmcgav:
@bryantbiggs I've updated this PR following the discussion we had on https://github.com/terraform-aws-modules/terraform-aws-iam/pull/479#issuecomment-2203812126

examples/iam-assumable-role-with-oidc/main.tf
```diff
@@ -89,3 +89,29 @@ module "iam_assumable_role_inline_policy" {
     }
   ]
 }
+
+#####################################
+# IAM assumable role with session tags
+#####################################
+module "iam_assumable_role_session_tags" {
```
bryantbiggs, line 96: we don't need a new example, just add to the existing

modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -43,7 +45,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
 
     content {
       effect  = "Allow"
-      actions = ["sts:AssumeRoleWithWebIdentity"]
```
bryantbiggs, line 46: leave this and just add `"sts:TagSessions"`

fatmcgav: So _always_ have the `sts:TagSessions` permission aswell?

modules/iam-assumable-role-with-oidc/main.tf
```diff
@@ -80,6 +82,16 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
           values   = var.oidc_fully_qualified_audiences
         }
       }
+
+      dynamic "condition" {
```
bryantbiggs, line 86: replace:

```hcl
  dynamic "condition" {
        for_each = var.provider_conditions

        content {
          test     = condition.value.test
          values   = condition.value.values
          variable = condition.value.variable
        }
      }
```

modules/iam-assumable-role-with-oidc/variables.tf
```diff
@@ -111,3 +111,15 @@ variable "allow_self_assume_role" {
   type        = bool
   default     = false
 }
+
+variable "enable_session_tags" {
```
bryantbiggs, line 115: replace with (I think this is right - should test this):

```hcl
variable "provider_conditions" {
  description = "[Condition constraints](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#condition) applied to the provider portion of the trust policy"
  type        = any
  default     = []
}

File: modules/iam-assumable-role-with-oidc/variables.tf [hunk 5]
```diff
@@ -111,3 +111,15 @@ variable "allow_self_assume_role" {
...
+  description = "Should the trust policy allow the `sts:TagSession` permission?"
+  type        = bool
+  default     = false
+}
+
+variable "oidc_session_tags" {
```
bryantbiggs, line 121: remove

modules/iam-assumable-role-with-oidc/main.tf
bryantbiggs: yes

fatmcgav:
@bryantbiggs OK, I've tweaked the PR to make it more generic. 
I did make one tweak to your suggestions, which was to call the variable `polcy_conditions` rather than `prvovider_conditions`, as I think the former is closer to the docs.

PTAL 👍

examples/iam-assumable-role-with-oidc/main.tf
```diff
@@ -89,3 +89,34 @@ module "iam_assumable_role_inline_policy" {
...
+
+  provider_url = "oidc.circleci.com/org/<CIRCLECI_ORG_UUID>"
+
+  oidc_fully_qualified_audiences = ["<CIRCLECI_ORG_UUID>"]
+
+  role_policy_arns = [
+    "arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+  ]
+
+  policy_conditions = [
```
bryantbiggs, line 115: we can add to existing example instead of creating new

modules/iam-assumable-role-with-oidc/main.tf
bryantbiggs: `policy_` is too generic - what policy is this referring to? This is only valid in the context of `var.provider_urls` or `var.provider_url` hence why I suggested `provider_conditions`. You could also do `provider_trust_policy_conditions` but just `policy_conditions` is too ambiguous for the current use

fatmcgav: Sure, renamed to `provider_trust_policy_conditions` 👍

fatmcgav:
Apologies, just noticed the failed pre-commit check... Have fixed that now so should be good 🤞

bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.44.0 :tada:

PR #484, opened 2024-05-08
feat: Add: output `external_dns_policy_arn` to module iam-role-for-service-accounts-eks

Diaa-Hassan:
## Description
This pull request adds a new output for the `external_dns_policy_arn`. This output provides the ARN of the IAM policy for external-dns. This information can be useful for other parts of the codebase that need to reference this policy.

## Motivation and Context
This can help in indexing and using this arn by other modules

## How Has This Been Tested?

### DIFF
--- a/examples/iam-role-for-service-accounts-eks/README.md
+++ b/examples/iam-role-for-service-accounts-eks/README.md
@@ -78,5 +78,5 @@ No inputs.
 | <a name="output_iam_role_name"></a> [iam\_role\_name](#output\_iam\_role\_name) | Name of IAM role |
 | <a name="output_iam_role_path"></a> [iam\_role\_path](#output\_iam\_role\_path) | Path of IAM role |
 | <a name="output_iam_role_unique_id"></a> [iam\_role\_unique\_id](#output\_iam\_role\_unique\_id) | Unique ID of IAM role |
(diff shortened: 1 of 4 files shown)


bryantbiggs:
> This information can be useful for other parts of the codebase that need to reference this policy

Such as? This sounds like you are leaking details across boundaries which will lead to conflicts or surprises

Diaa-Hassan:
@bryantbiggs 
how would this lead to conflicts and surprises ??!!!

```
resource "aws_iam_role_policy_attachment" "example" {
  for_each   = module.cluster.eks_managed_node_groups
  policy_arn = module.eks_external_dns_iam.external_dns_policy_arn
  role       = each.value.iam_role_name
}
```

how can the policy be attached to any role if the policy name and arn not known ???
the policy generated has by default the prefix `AmazonEKS_External_DNS_Policy-` and followed by a random number that can't be retrieved unless there is an output of this generated policy arn

bryantbiggs:
well:
1. You don't attach the external DNS policy to the node IAM role, thats an anti-pattern
2. If you rely on the node IAM role for providing the permissions, other applications will inherit those permissions. If you later remove this from the node IAM role and use it only on the external DNS chart (namespace/service account), then you could break those other applications that had inherited permissions that were not known

In short, you use the IAM role that is generated from this module, not its policy

PR #488, opened 2024-05-13
feat: Add dynamic block to support job_workflow_ref

benchunghpe:
## Description
This PR adds a dynamic block and optional variable to support specifying a set of allowed `job_workflow_ref` to allow fine-grained access to a github OIDC role.

## Motivation and Context
Currently, the github OIDC role module doesn't support passing `job_workflow_ref` to explicitly allow a limited set of workflows to assume an AWS IAM role through OIDC. The potential impact of this is that using this module, someone would be able to fork one of our github workflows, change the business logic and still have no issues assuming the IAM role.

At HPE, we have a local version of this module which allows us to say "only allow a workflow to assume this IAM role if it's coming from the main branch of our organisation-wide reusable-workflows", which isn't possible in the main branch of this module due to the lack of support.

## Breaking Changes
n/a, this adds a new optional variable so should extend the current functionality for those who require this feature

## How Has This Been Tested?
- [x] This is running in production at HPE
- [x] I have executed pre-commit run -a on my pull request

### DIFF
--- a/examples/iam-github-oidc/README.md
+++ b/examples/iam-github-oidc/README.md
@@ -35,6 +35,7 @@ Run `terraform destroy` when you don't need these resources.
 
 | Name | Source | Version |
 |------|--------|---------|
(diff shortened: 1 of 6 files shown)


benchunghpe:
This should be ready for review when someone's available.

antonbabenko:

Issue #500, opened 2024-07-22
CircleCI OIDC provider and role


simonweil:
## Describe the solution you'd like.

Would you accept a pull request implementing CircleCI OIDC provider and role like for GHA?

## Describe alternatives you've considered.

Keeping it to myself, rather than sharing with the community :)

simonweil:
As the PR is open, should we keep this one?

antonbabenko:
We can have a PR for the generic support and close this issue.

PR #501, opened 2024-07-24
feat(circleci-oidc): Add support for the provider and the role

simonweil:
## Description

Created a CircleCI OIDC provider and role

## Motivation and Context

It's hard to do it on your own and I'm happy to contribute it to the community

## Breaking Changes

None

## How Has This Been Tested?

Closes: #500

### DIFF
--- a/modules/iam-circleci-oidc-provider/variables.tf
+++ b/modules/iam-circleci-oidc-provider/variables.tf
@@ -1,0 +1,33 @@
+variable "create" {
+  description = "Controls if resources should be created (affects all resources)"
+  type        = bool
+  default     = true
+  nullable    = false
+}
+
+variable "tags" {
+  description = "A map of tags to add to the resources created"
+  type        = map(any)
+  default     = {}
+  nullable    = false
+}
+
+variable "org_uuid" {
+  description = "The CircleCI organization UUID that will be authorized to assume the role."
+  type        = string
+  nullable    = false
+}
+
+variable "base_url" {
+  description = "The URL of the identity provider. Corresponds to the iss claim"
+  type        = string
+  default     = "https://oidc.circleci.com/org"
+  nullable    = false
+}
+
+variable "thumbprints" {
+  description = "List of thumbprints of CircleCI"
+  type        = list(string)
+  default     = ["9e99a48a9960b14926bb7f3b02e22da2b0ab7280"]
+  nullable    = false
+}
--- a/modules/iam-circleci-oidc-role/main.tf
+++ b/modules/iam-circleci-oidc-role/main.tf
@@ -1,0 +1,55 @@
+locals {
+  # Just extra safety incase someone passes in a url with `https://`
+  base_url = replace(var.base_url, "https://", "")
+}
+
+################################################################################
+# CircelCI OIDC Role
+################################################################################
+
+data "aws_iam_policy_document" "this" {
+  count = var.create ? 1 : 0
+
+  statement {
+    sid    = "CircleCIOidcAuth"
+    effect = "Allow"
+    actions = [
+      "sts:TagSession",
+      "sts:AssumeRoleWithWebIdentity",
+    ]
+
+    principals {
+      type        = "Federated"
+      identifiers = [var.oidc_provider_arn]
+    }
+
+    condition {
+      test     = "StringLike"
+      variable = "${local.base_url}/${var.org_uuid}:sub"
+      values   = [for project_uuid in var.project_uuids : "org/${var.org_uuid}/project/${project_uuid}/user/*"]
+    }
+  }
+}
+
+resource "aws_iam_role" "this" {
+  count = var.create ? 1 : 0
+
+  name        = var.name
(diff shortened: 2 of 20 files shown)


simonweil:
@antonbabenko why does CI fail on `nullable    = false` in variables?
This is very useful to make sure defaults are used and unexpected `null`s are not passed to the module.

simonweil:
@antonbabenko ping...

simonweil:
@bryantbiggs @antonbabenko any chance to get a go or no go?
I'm using my version of this module now for a around a month and all is good.

modules/iam-circleci-oidc-provider/variables.tf
```diff
@@ -0,0 +1,33 @@
...
+  description = "A map of tags to add to the resources created"
+  type        = map(any)
+  default     = {}
+  nullable    = false
+}
+
+variable "org_uuid" {
+  description = "The CircleCI organization UUID that will be authorized to assume the role."
+  type        = string
+  nullable    = false
```
antonbabenko, line 18: When `create = false`, the variables without default values must still be set since nullable is set to `false`. Right?

https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/examples/iam-github-oidc/main.tf#L57-L61

---

Please add examples for this module similar to https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/examples/iam-github-oidc/

modules/iam-circleci-oidc-role/main.tf
```diff
@@ -0,0 +1,55 @@
+locals {
+  # Just extra safety incase someone passes in a url with `https://`
```
antonbabenko, line 2:
```suggestion
  # Just extra safety in case someone passes in a url with `https://`
```

modules/iam-circleci-oidc-role/variables.tf
```diff
@@ -0,0 +1,92 @@
...
+  description = "IAM role name prefix"
+  type        = string
+  default     = null
+}
+
+variable "policies" {
+  description = "Policies to attach to the IAM role in `{'static_name' = 'policy_arn'}` format"
+  type        = map(string)
+  default     = {}
+  nullable    = false
```
antonbabenko, line 54: Nullables are not needed in most places in the module. At least, I think so. We don't have them in pretty much any modules.

@bryantbiggs What do you think?

antonbabenko:
`nullable` requires terraform >= 1.1.0 but the module you add requires 1.0+.

If we decide that we need nullable, we need to update `versions.tf` to get pre-commit green. @bryantbiggs WDYT?

antonbabenko (changes requested):

modules/iam-circleci-oidc-provider/variables.tf
simonweil: Yes, thinking about it, what is the point in the `create` variable?
Will add the example.

modules/iam-circleci-oidc-role/variables.tf
simonweil: IMHO, it is very useful to set `nullable = false` anywhere you do not expect to get `null`s so you do not need to handle situations when `null` is set and there is an expected default. Also it is useful in sub modules for handling defaults by setting null in the input to make that default of the variable take place.

We do it for all our modules :)

But, your call, I can remove if you think it is better for the community.

bryantbiggs: I don't see the value in using `nullable` - we haven't seen a use case yet where the resolution was to use `nullable`

bryantbiggs:
in general - I don't like the idea of having a module per CI/CD provider. I know in the past I have pushed back when someone wanted to add a new module for Bamboo CI. 

If we went forward, I would rather we create a module that is generic, and lets users opt into the different 3rd party provider (default) functionality. Something like `iam-3rd-party-oidc-role` or `iam-integration-oidc-role` and then we can support various integrations such as GitHub, GitHub Enterprise, Bamboo CI, CircleCI, etc. @antonbabenko what are your thoughts

antonbabenko:
I agree that we don't want to have a module per CI/CD provider, so let's update this PR to bring generic `iam-integration-oidc-role` and have a set of predifined properties per provider and a way to specify custom ones. It is also important to show the usages in examples.

In the longer run, we may discontinue the one we already have for GitHub.

simonweil:
> I agree that we don't want to have a module per CI/CD provider, so let's update this PR to bring generic `iam-integration-oidc-role` and have a set of predifined properties per provider and a way to specify custom ones. It is also important to show the usages in examples.
> 
> In the longer run, we may discontinue the one we already have for GitHub.

Sounds good to me, I'll sketch up a direction for discussion

modules/iam-circleci-oidc-role/variables.tf
simonweil: Ok

simonweil:
Don't close, I'm planning to get to it

PR #515, opened 2024-09-08
fix!: Irsa additional assumable principals

sandeshgrangdan:
## Description

Enhanced the `iam-role-for-service-accounts-eks` module to support additional trusted principals including both IAM roles and users.

## Motivation and Context

To allow an IRSA role from one service to be assumed by an IRSA role in another service, supporting cross-service role assumption.

## Breaking Changes

None

## How Has This Been Tested?


sandeshgrangdan:
> > To allow an IRSA role from one service to be assumed by an IRSA role in another service, supporting cross-service role assumption.
> 
> Why?

I need to generate temporary credentials from an IRSA role that will be used by external applications. To manage these credentials, I have a separate key rotation application with its own role or user that will update the temporary credentials generated by the IRSA role.

bryantbiggs:
thank you for that info - unfortunately, I don't think that is a use case we are going to support here

PR #551, opened 2025-02-17
feat: Support inline policies for EKS IRSA

aleskiontherun:
## Description

This commit adds support for creating an inline policy for an IAM role to iam-role-for-service-accounts-eks module.

Inspired by https://github.com/terraform-aws-modules/terraform-aws-iam/pull/479

## Motivation and Context

It is often useful to create custom inline policies in place, especially useful with Terragrunt where otherwise a custom policy would need to be created in a dedicated root module.

## Breaking Changes

Nope.

## How Has This Been Tested?

### DIFF
--- a/examples/iam-role-for-service-accounts-eks/README.md
+++ b/examples/iam-role-for-service-accounts-eks/README.md
@@ -49,6 +49,7 @@ Run `terraform destroy` when you don't need these resources.
 | <a name="module_iam_eks_role"></a> [iam\_eks\_role](#module\_iam\_eks\_role) | terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts-eks | n/a |
 | <a name="module_iam_policy"></a> [iam\_policy](#module\_iam\_policy) | terraform-aws-modules/iam/aws//modules/iam-policy | n/a |
 | <a name="module_irsa_role"></a> [irsa\_role](#module\_irsa\_role) | ../../modules/iam-role-for-service-accounts-eks | n/a |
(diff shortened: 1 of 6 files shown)


bryantbiggs:
thank you - unfortunately that would defeat the purpose of the canned policies so we won't be accepting the PR as is

aleskiontherun:
@bryantbiggs could you elaborate a bit or point me to some more information about your decision? I don't quite understand why is it OK to have inline policies in other role modules, but not in IRSA...

bryantbiggs:
the other modules are generic in terms of what policy is used - the IRSA module is not. there are two key aspects of the IRSA module:
1. The assume role trust policy for IRSA
2. The canned policies for the different projects that required IAM permissions, typically provided through either IRSA or now Pod Identity

aleskiontherun:
@bryantbiggs The canned policies are good for very specific tools in very specific use-cases. E.g. I can deploy cert-manager app to EKS and easily create a pre-configured role for it, that's great. But you can't have canned roles for all the apps in the world, that's kind of insane. Now I have my own app that needs specific access to AWS and using an IRSA is a very straight forward decision. From my perspective, an IRSA is not different from a regular IAM role, except it can be attached to a K8s service account. AWS allows creating inline policies for IRSAs, I can also easily do it with Terraform. But I'm using Terragrunt and without inline policy support by the module, I would have to create a dedicated module just to create a standalone policy that is only needed once and will never be reused, and then add it as a dependency to my role module, which adds unnecessary complexity. So not supporting inline policies in the module just complicates things and doesn't add any value IMO.

antonbabenko:

PR #561, opened 2025-04-01
fix: IAM policy for External-DNS policy

vmignot:
## Description
IAM policy for External-DNS has changed.

`route53:ListTagsForResource` => `route53:ListTagsForResource`**s**

See https://github.com/kubernetes-sigs/external-dns/commit/658d1efb76d7ce1f454cec9e8e92192c6c19d6fb

## Motivation and Context
The policy does not work with recent external-dns version. It aligns the policy with the doc

## Breaking Changes
The policy will not work with older version. We could provide backward compatibility by either providing both actions `ListTagsForResources` and `ListTagsForResource`, or a toggle.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/policies.tf
+++ b/modules/iam-role-for-service-accounts-eks/policies.tf
@@ -501,7 +501,7 @@ data "aws_iam_policy_document" "external_dns" {
     actions = [
       "route53:ListHostedZones",
       "route53:ListResourceRecordSets",
-      "route53:ListTagsForResource",
+      "route53:ListTagsForResources",
     ]
 
     resources = ["*"]


bryantbiggs (approved):

antonbabenko:
This PR is included in version 5.54.1 :tada:

octane100:
It seems that this commit contains a breaking change. A minor version would have been welcome if it's really what was intented.

Furthermore, why was the old `route53:ListTagsForResource` not simply kept ? This way, old versions of external-dns would continue to work seamlessly. Am I missing something ?

bryantbiggs:
we are matching the upstream policy. if you wish to keep the prior version, you can pin your module version (which is always recommended)

PR #570, opened 2025-05-29
chore: Use appropriate mapping of assume_role_policy

waskow-consensys:
## Description
In
https://aws.amazon.com/blogs/security/announcing-an-update-to-iam-role-trust-policy-behavior/
we need to allow the role to assume itself.
Current configuration was set to `AWS: "*"` with a conditional.
This is detected as a bad policy by AWS, where it flags the trust policy

>Overly permissive trust policy exists in your trust relationships
Broad access: Principals that include a wildcard (*, ?) can be overly permissive.

Additionally this "Self Assume" causes IAM looping which can lead to multiple requests, increase cloudtrail audit logs, etc

## Motivation and Context
This saves security teams from seeing `AWS: "*"` and panicking 

## Breaking Changes
No

## How Has This Been Tested?

### DIFF
--- a/modules/iam-assumable-role-with-oidc/main.tf
+++ b/modules/iam-assumable-role-with-oidc/main.tf
@@ -27,13 +27,7 @@ data "aws_iam_policy_document" "assume_role_with_oidc" {
 
       principals {
         type        = "AWS"
-        identifiers = ["*"]
-      }
-
-      condition {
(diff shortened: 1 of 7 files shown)


bryantbiggs:
it is correct as written per the AWS docs, closing since this is not an appropriate change

antonbabenko:

PR #576, opened 2025-07-03
fix: Pin Terraform version to v1.12.2 and simplify GitHub Actions workflow

Kkasuga904:
## Overview
- Pinned Terraform version to `v1.12.2` for stability.
- Removed dynamic version extraction using `terraform-min-max`.
- Simplified the workflow by eliminating unnecessary matrix and condition logic.
- Improved maintainability and reliability of the pre-commit workflow.

## Background
The previous setup dynamically determined the min/max Terraform versions, which introduced instability when new versions were released.  
By explicitly setting the version to `v1.12.2` (the latest stable release as of July 2025), we ensure consistent CI behavior.

## Affected Areas
- Only `.github/workflows/pre-commit.yml` is modified.
- No changes to Terraform code itself.

## Test Plan
- Confirm that the GitHub Actions workflow passes with the pinned Terraform version.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -7,6 +7,7 @@ on:
       - master
 
 env:
+  TERRAFORM_VERSION: 1.12.2
   TERRAFORM_DOCS_VERSION: v0.19.0
   TFLINT_VERSION: v0.53.0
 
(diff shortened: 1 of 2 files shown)


bryantbiggs:
I don't think you understand what the CI checks are testing - dynamically checking the Terraform version bounds is intentional

Kkasuga904:
Thanks for the explanation! I understand now that it was intentional. Appreciate you taking the time to clarify

PR #581, opened 2025-08-07
fix: Replace deprecated aws_region data source name attribute

atilsensalduz:
Replaces the deprecated `data.aws_region.current.name` with `data.aws_region.current.region` in the IAM Role for Service Accounts EKS module.

## Description
- Updated `modules/iam-role-for-service-accounts-eks/main.tf` to use the non-deprecated `region` attribute instead of `name`
- 
## Motivation and Context
The `name` attribute of the `aws_region` data source has been deprecated by the AWS provider. This change ensures compatibility with current and future versions of the AWS provider.

### DIFF
--- a/modules/iam-role-for-service-accounts-eks/main.tf
+++ b/modules/iam-role-for-service-accounts-eks/main.tf
@@ -6,7 +6,7 @@ locals {
   account_id          = data.aws_caller_identity.current.account_id
   partition           = data.aws_partition.current.partition
   dns_suffix          = data.aws_partition.current.dns_suffix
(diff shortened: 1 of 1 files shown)


bryantbiggs:
I think theres a few more changes we'll want to make since its a breaking change. its on the list and we'll get to it soon

atilsensalduz:
thank you! @bryantbiggs I'm also interested in contributing to upgrades for the remaining modules.

bryantbiggs:
thank you but this upgrade cycle is a bit more involved for the modules hosted here - you can see more at #585

Issue #586, opened 2025-08-13
ambiguous moved statements in iam-role-for-service-accounts/migrations.tf


jordanb-afs:
Recently tried upgrading to 6.0.0 version of this module and immediately ran into this issue. it seems multiple moved statements are all pointing to the same thing:

```
│ Error: Ambiguous move statements
│
│   on .terraform/modules/eks_cluster.ebs_csi_irsa_role/modules/iam-role-for-service-accounts/migrations.tf line 160:
│  160: moved {
│
│ A statement at .terraform/modules/eks_cluster.ebs_csi_irsa_role/modules/iam-role-for-service-accounts/migrations.tf:6,1 declared that
│ module.eks_cluster[0].module.ebs_csi_irsa_role.aws_iam_policy.aws_gateway_controller moved to module.eks_cluster[0].module.ebs_csi_irsa_role.aws_iam_policy.this, but this statement instead declares that
│ module.eks_cluster[0].module.ebs_csi_irsa_role.aws_iam_policy.vpc_cni moved there.
│
│ Each resource can have moved from only one source resource.
╵
╷
│ Error: Ambiguous move statements
│
│   on .terraform/modules/eks_cluster.ebs_csi_irsa_role/modules/iam-role-for-service-accounts/migrations.tf line 165:
│  165: moved {
│
│ A statement at .terraform/modules/eks_cluster.ebs_csi_irsa_role/modules/iam-role-for-service-accounts/migrations.tf:6,1 declared that
│ module.eks_cluster[0].module.ebs_csi_irsa_role.aws_iam_policy.aws_gateway_controller moved to module.eks_cluster[0].module.ebs_csi_irsa_role.aws_iam_policy.this, but this statement instead declares that
│ module.eks_cluster[0].module.ebs_csi_irsa_role.aws_iam_role_policy_attachment.vpc_cni moved there.
│
│ Each resource can have moved from only one source resource.
╵
```

bryantbiggs:
We'll need a reproduction in order to assist

bryantbiggs:
hmmm, ok - not sure why that didn't get picked up. I might get to this later tonight or worst case tomorrow morning

kaiyuanlim-adh:
We also have the same issue as well.

lukaspj:
Previously this module would create one IAM Policy per "function"
I.e. one for Velero, one for VPC CNI etc.

[...]

PR #588, opened 2025-08-14
fix: Correct types in 6.0 moved blocks mismatch between aws_iam_policy and aws_iam_role_policy_attachment

lukaspj:
## Description

There seems to have been a copy-paste mistake in the migrations where `aws_iam_policy` type has been used as the "to" for `aws_iam_role_policy_attachment` resources.

## Motivation and Context

We are seeing the following error:
```
│ Error: Move Resource State Not Supported
│ 
│ The "aws_iam_policy" resource type does not support moving resource state
│ across resource types.
```
Edit: This does not solve #586 as that is related to the change of merging all the policies into a single resource being a breaking change

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role-for-service-accounts/migrations.tf
+++ b/modules/iam-role-for-service-accounts/migrations.tf
@@ -10,7 +10,7 @@ moved {
 
 moved {
   from = aws_iam_role_policy_attachment.aws_gateway_controller
-  to   = aws_iam_policy.this
+  to   = aws_iam_role_policy_attachment.this
 }
 
 # Cert Manager
@@ -21,7 +21,7 @@ moved {
 
 moved {
   from = aws_iam_role_policy_attachment.cert_manager
-  to   = aws_iam_policy.this
+  to   = aws_iam_role_policy_attachment.this
 }
 
 # Cluster Autoscaler
@@ -32,7 +32,7 @@ moved {
 
 moved {
   from = aws_iam_role_policy_attachment.cluster_autoscaler
-  to   = aws_iam_policy.this
+  to   = aws_iam_role_policy_attachment.this
 }
 
 # EBS CSI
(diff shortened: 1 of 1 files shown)


botchk:
Appreciate the quick response. Just ran into this issue and tried to fix it locally But the proposed fix doesn't solve the underlying issue of multiple different resources trying to move to the same resource, which is something that Terraform doesn't like:

```
Error: Ambiguous move statements
 
   on .terraform/modules/cluster_autoscaler_irsa/modules/iam-role-for-service-accounts/migrations.tf line 165:
  165: moved {
 
A statement at .terraform/modules/cluster_autoscaler_irsa/modules/iam-role-for-service-accounts/migrations.tf:11,1 declared hat module.cluster_autoscaler_irsa.aws_iam_role_policy_attachment.aws_gateway_controller
moved to module.cluster_autoscaler_irsa.aws_iam_role_policy_attachment.this, but this statement instead declares that module.cluster_autoscaler_irsa.aws_iam_role_policy_attachment.vpc_cni moved there.

Each resource can have moved from only one source resource.

```

I don't know that the intended migration scenario was but looks like the migration concept needs to be revisited.

lukaspj:
Ah right of course, and that's also why we haven't been encountering your issue because we only invoke this module once per policy.

I'd say isn't the best solution here to destroy the existing policies and recreate a new one instead of trying to migrate it?

botchk:
> I'd say isn't the best solution here to destroy the existing policies and recreate a new one instead of trying to migrate it?

Yes, I also agree that this is the best approach. Consumers can then still decide if they want to do a migration and add the relevant moved blocks themselves.

lukaspj:
Tried destroying existing policies and recreating with the module as-is and got this error:

image

So I'd reckon this PR doesn't properly solve the problem. The only solution, as I see it, is to delete the migrations and force them to recreate their IAM Policies or write their own migrations..

PR #595, opened 2025-08-14
fix: Attach policies to IAM user (not role)

emuesuenu:
## Description
Replace the incorrect resource type aws_iam_role_policy_attachment with aws_iam_user_policy_attachment.

Update arguments accordingly (role -> user) so attachments target the created IAM user (aws_iam_user.this[0].name).

No changes to input variables: var.policies remains a map and the for_each logic is unchanged.

## Motivation and Context
The previous implementation attempted to attach policies using the role attachment resource for an IAM user, which is invalid.

## Breaking Changes
Resource address change: from aws_iam_role_policy_attachment.additional to aws_iam_user_policy_attachment.additional.
Terraform will plan to replace these attachment resources on the next apply (destroy old, create new).

Impact: Brief re‑creation of policy attachments during apply. No persistent permission loss is expected.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-user/README.md
+++ b/modules/iam-user/README.md
@@ -50,9 +50,9 @@ No modules.
 | Name | Type |
 |------|------|
 | [aws_iam_access_key.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_access_key) | resource |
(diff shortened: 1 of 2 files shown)


bryantbiggs:
This was corrected in #594

PR #631, opened 2025-12-12
docs(oidc-provider-example): Update requirements and providers section

lanandra:
## Description

Update oidc provider example README with tls requirements

## Motivation and Context

When i tried Terraform Kiro powers, i found Kiro said that oidc-provider submodule don't need tls provider. I realized in my GitHub Actions, terraform init use that tls provider

Some words from Kiro

> You're absolutely right - that was my mistake! I should have looked at the actual module implementation, not just the examples.
> 
> The examples show "No providers" because they're just calling the modules, but the actual submodule code (in the modules/ directory) is what matters - that's where the real provider requirements and data sources are defined.
> 
> The examples directory shows how to use the modules, but the modules directory contains the actual implementation with:
> 
>  - The real required_providers block
> - Data sources like data.tls_certificate
> - The actual resources being created
>
> When the Terraform Power returned the module documentation, it gave me the examples section which was misleading. I should have:
> 
> 1.  Asked for more detailed module information
> 2. Checked the actual submodule source code
> 3. Or better yet - looked at what terraform init was actually downloading and requiring (like your GitHub Actions log showed)
>
> Your GitHub Actions output was the truth - it clearly showed:
> 
> 
```
- Finding hashicorp/tls versions matching ">= 3.0.0"...
module.github_oidc_provider.data.tls_certificate.this[0]: Reading...
```
>
> That's the actual module behavior, not what the example documentation suggested. Good catch! This is why checking actual Terraform output and looking at the real module source is more reliable than just reading example documentation.

For some consistency and feeding correct information such as to AI agent, i think we can make the README in examples similar to what written in modules README

## Breaking Changes

N/A

## How Has This Been Tested?

### DIFF
--- a/examples/iam-oidc-provider/README.md
+++ b/examples/iam-oidc-provider/README.md
@@ -30,10 +30,7 @@ Run `terraform destroy` when you don't need these resources.
 
 ## Providers
 
-| Name | Version |
-|------|---------|
-| <a name="provider_aws"></a> [aws](#provider\_aws) | >= 6.0 |
-| <a name="provider_tls"></a> [tls](#provider\_tls) | >= 3.0 |
+No providers.
 
 ## Modules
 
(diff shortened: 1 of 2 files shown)


bryantbiggs:
this is incorrect

PR #636, opened 2026-02-15
feat: Allow setting cert thumbprints manually in OIDC provider

k-k-k-k-k-k-k-k:
## Description

With this change you can manually set cert thumbprints of an OIDC provider.

## Motivation and Context

In order to prevent service disruptions, you have to to add cert thumbprints before actually switching certs, at least when you use certs that do not chain to certs in IAM's internal trust store. Apart from that, using the data source `tls_certificate` to determine cert thumbprints can be problematic because the server/intermediate/root cert will usually change from time to time thus leading to a diff or because the machine running _terraform apply_ might not be able to reach the OIDC provider or only through an SSL termination proxy.

## Breaking Changes

None.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-oidc-provider/main.tf
+++ b/modules/iam-oidc-provider/main.tf
@@ -11,7 +11,7 @@ locals {
 ################################################################################
 
 data "tls_certificate" "this" {
-  count = var.create ? 1 : 0
+  count = var.create && length(var.thumbprint_list) == 0 ? 1 : 0
 
   url = var.url
 }
@@ -21,7 +21,7 @@ resource "aws_iam_openid_connect_provider" "this" {
 
   url             = var.url
   client_id_list  = coalescelist(var.client_id_list, ["sts.${local.dns_suffix}"])
-  thumbprint_list = data.tls_certificate.this[0].certificates[*].sha1_fingerprint
+  thumbprint_list = coalescelist(var.thumbprint_list, try(data.tls_certificate.this[0].certificates[*].sha1_fingerprint, []))
 
   tags = var.tags
 }
--- a/modules/iam-oidc-provider/variables.tf
+++ b/modules/iam-oidc-provider/variables.tf
@@ -20,6 +20,12 @@ variable "client_id_list" {
   default     = []
 }
 
+variable "thumbprint_list" {
+  description = "A list of server certificate thumbprints (max 5) for the OpenID Connect (OIDC) identity provider’s server certificates"
+  type        = list(string)
+  default     = []
+}
+
 variable "url" {
   description = "The URL of the identity provider. Corresponds to the iss claim"
   type        = string
--- a/examples/iam-oidc-provider/README.md
+++ b/examples/iam-oidc-provider/README.md
@@ -38,6 +38,8 @@ No providers.
 | <a name="module_github_oidc_iam_provider"></a> [github\_oidc\_iam\_provider](#module\_github\_oidc\_iam\_provider) | ../../modules/iam-oidc-provider | n/a |
 | <a name="module_github_oidc_iam_role"></a> [github\_oidc\_iam\_role](#module\_github\_oidc\_iam\_role) | ../../modules/iam-role | n/a |
 | <a name="module_oidc_iam_provider_disabled"></a> [oidc\_iam\_provider\_disabled](#module\_oidc\_iam\_provider\_disabled) | ../../modules/iam-oidc-provider | n/a |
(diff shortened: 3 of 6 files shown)


k-k-k-k-k-k-k-k:
@antonbabenko, @bryantbiggs, you have reviewed the last couple pull requests in this repo, could you take a look at this one when you get a chance? Many thanks in advance.

modules/iam-oidc-provider/main.tf
```diff
@@ -21,7 +21,7 @@ resource "aws_iam_openid_connect_provider" "this" {
 
   url             = var.url
   client_id_list  = coalescelist(var.client_id_list, ["sts.${local.dns_suffix}"])
-  thumbprint_list = data.tls_certificate.this[0].certificates[*].sha1_fingerprint
+  thumbprint_list = var.thumbprint_list == [] ? data.tls_certificate.this[0].certificates[*].sha1_fingerprint : var.thumbprint_list
```
bryantbiggs:
```suggestion
  thumbprint_list = coalescelist(var.thumbprint_list, data.tls_certificate.this[0].certificates[*].sha1_fingerprint)
```

modules/iam-oidc-provider/variables.tf
```diff
@@ -25,3 +25,9 @@ variable "url" {
   type        = string
   default     = "https://token.actions.githubusercontent.com"
 }
+
+variable "thumbprint_list" {
+  description = "List of up to 5 server certificate thumbprints used to verify the connection with the identity provider url in case the server certificate is signed by a root CA that is not in IAM's builtin trust store. When renewing the identity provider server certificate, ideally add its thumbprint before the switch"
```
bryantbiggs:
```suggestion
  description = "A list of server certificate thumbprints (max 5) for the OpenID Connect (OIDC) identity provider’s server certificates"
```

modules/iam-oidc-provider/main.tf
k-k-k-k-k-k-k-k: Thanks for the suggestion but unfortunately this does not work when setting `thumbprint_list = null` which is a use case I would like to cover.

k-k-k-k-k-k-k-k: I just realized that my conditional expression is flawed. Going to fix it.

bryantbiggs: thats never going to be a valid use case we support in the implementation though

modules/iam-oidc-provider/variables.tf
k-k-k-k-k-k-k-k: Changed it, thanks.

modules/iam-oidc-provider/main.tf
k-k-k-k-k-k-k-k: I see. I have reworked the pull request and hope this is the right approach. Let me know what you think.

bryantbiggs (changes requested):
this is a breaking change and not in a direction that we want to go so we won't be accepting it

k-k-k-k-k-k-k-k:
> this is a breaking change and not in a direction that we want to go so we won't be accepting it

What would be the best approach in order to allow setting server cert thumbprints manually as well as setting none at all (`null`)?

bryantbiggs:
the approach I proposed above with coalescelist is the correct approach for manually setting thumbprints - I don't see any valid scenario where you would not set a thumbprint, that doesn't make sense at all

k-k-k-k-k-k-k-k:
> the approach I proposed above with coalescelist is the correct approach for manually setting thumbprints - I don't see any valid scenario where you would not set a thumbprint, that doesn't make sense at all

Thanks for the feedback. To quote hashicorp/terraform-provider-aws/pull/37255: AWS now secures communication with some OIDC identity providers (IdPs) through a library of trusted root certificate authorities (CAs) instead of using a certificate thumbprint to verify the IdP server certificate.

That's why the argument `thumbprint_list` became optional thus it does make sense to not set a thumbprint, imho.

I have updated the pull request as requested anyway. Hope this works for you.

bryantbiggs:
maybe - but per what you linked https://github.com/hashicorp/terraform-provider-aws/issues/40509 

I think we're just spinning around and around trying to solve an xy-problem. so for now, I'm inclined to simply close this out and not waste anymore time

k-k-k-k-k-k-k-k:
Sorry that my explanation of the issue and my attempts to solve it did not satisfy you. I am not a native speaker but I tried my best. I hope you will reconsider one day.

Issue #285, opened 2022-10-06
AWS provides a managed policy for EBS CSI drive


obounaim:
## Is your request related to a new offering from AWS?
I am not sure if it is a new offering form AWS, but I would say yes. Maybe the managed policy did not exists when this module was first created.

  - Yes ✅: not related to the terraform AWS provider

## Is your request related to a problem? Please describe.

This issue concerns the module iam-role-for-service-accounts-eks. When using this module to create an IAM role for EBS CSI driver, it creates a new customer managed IAM policy. This should be fine, however I just wanted to point out that AWS is already providing a managed policy for EBS driver (maybe you already aware of this, and there is a good reason to not use it).

Managed policy ARN: arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy

## Describe the solution you'd like.
Just attach this managed policy using its ARN to the new role without creating a new IAM policy

## Additional context
[AWS Documentation](https://docs.aws.amazon.com/eks/latest/userguide/csi-iam-role.html)

bryantbiggs:
While that does work for most cases, I don't believe that policy supports encrypted volumes if I remember correctly https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts-eks/policies.tf#L251

obounaim:
True the managed policy does not give KMS rights.

However, I would still use the AWS policy for basic rights needed by the EBS CSI drive and for KMS I would add a variable attach_ebs_csi_kms_policy to attach a second policy for KMS if needed by the user of the module. 

I think for a lot of use cases, people does not use customer KMS keys to encrypt EBS volumes, that why I would make an option to have KMS rights.

Like this we can achieve :
1. Use a managed policy and make AWS responsible for it
2. Implement least-privilege principal

At least this how I think about it, please let me know what is your opinion about it.

I can make a PR for it if you are OK with it, however this will be a major change for the module.

PR #640, opened 2026-04-23
chore: Replace inline EBS CSI policy with AWS-managed one

Roberdvs:
## Description
This PR replaces the inline policy for the EBS CSI driver with a reference to the AWS-managed [AmazonEBSCSIDriverPolicy](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/AmazonEBSCSIDriverPolicy.html) ARN while keeping the KMS statements that are missing from the upstream policy.

## Motivation and Context
The `AmazonEBSCSIDriverPolicy` policy is maintained by AWS and automatically updated with new permissions as the driver evolves, like it's doing recently with [`DescribeInstanceTypes`](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/CHANGELOG.md#describeinstancetypes-at-runtime). 

Instead of having to mirror these changes in the module codebase, this lets the role pick up new permissions without a module release.

This was already briefly discussed years ago in https://github.com/terraform-aws-modules/terraform-aws-iam/issues/285 but nothing came out of it.

## Breaking Changes

`AmazonEBSCSIDriverPolicy` is a superset of the inline policy apart from the KMS actions that are kept inline, so no breaking changes in the role itself, but this will destroy existing `aws_iam_policy` and `aws_iam_role_policy_attachment` if no KMS ARNs are set, so maybe those could be considered breaking changes.

## How Has This Been Tested?


bryantbiggs:
thank you but we are not planning to make this type of change - there are a myriad of reasons why we specify the same policy statements as a managed policy and thats usually intentional

Roberdvs:
Understood.

I'll wait for the release that includes the policy updates, then.

Cheers!

PR #650, opened 2026-06-23
fix: Handle null role name with name prefix

chrispsheehan:
## Description
This change updates modules/iam-role to handle `name = null` when `use_name_prefix = true`.

The module now derives name_prefix from a local value that is only set when var.name is non-null, instead of interpolating var.name directly. This avoids passing an invalid prefix to IAM resources.

## Motivation and Context
This fixes a Terraform plan failure caused by interpolating `null` into `name_prefix`:
```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: Invalid template interpolation value
│ 
│   on modules/iam-role/main.tf line 288, in resource "aws_iam_role" "this":
│  288:   name_prefix = var.use_name_prefix ? "${var.name}-" : null
│     ├────────────────
│     │ var.name is null
│ 
│ The expression result is null. Cannot include a null value in a string template.
```

## Breaking Changes
None.

## How Has This Been Tested?

### DIFF
--- a/modules/iam-role/main.tf
+++ b/modules/iam-role/main.tf
@@ -11,6 +11,7 @@ locals {
 
   oidc_providers     = [for url in var.oidc_provider_urls : replace(url, "https://", "")]
   bitbucket_provider = try(element(local.oidc_providers, 0), null)
+  name_prefix        = var.name != null ? "${var.name}-" : null
 }
 
 ################################################################################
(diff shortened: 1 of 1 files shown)


bryantbiggs:
I'm not following - lets start with an issue and a reproduction

PR #651, opened 2026-06-24
feat: Add ebs_csi_volume_tagging to allow EBS CSI tag modification on existing volumes

### LINKED ISSUE #630: EBS CSI Policy doesn't works with VolumeAttributesClass tagging
## Description

I searched for issues and found https://github.com/terraform-aws-modules/terraform-aws-iam/issues/562
But while using `6.2.3` version of module with EBS CSI Driver I'm still not able to use to modify volume using VolumeAttributesClass:
```
type="Warning" reason="VolumeModifyFailed" message="rpc error: code = Internal desc = Could not modify volume tags "vol-xx": operation error EC2: CreateTags, https response error StatusCode: 403, RequestID: id, api error UnauthorizedOperation: You are not authorized to perform this operation. User: <user> is not authorized to perform: ec2:CreateTags on resource: <resource> because no identity-based policy allows the ec2:CreateTags action.
```

Completely removing this [condition](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts/policies.tf#L219-L227) allows driver to tag existing volumes.

Example policy for using VAC [doens't have any conditions](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/docs/tagging.md#adding-modifying-and-deleting-tags-of-existing-volumes)

PS I also filled issue in driver repository to update example policy https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2799

## Versions

- Module version [Required]: 6.2.3
- Terraform version: Terraform v1.12.2
- Provider version(s): provider registry.terraform.io/hashicorp/aws v6.15.0

- Deploy EBS CSI Driver IRSA policy using this module
- Try to modify existing EBS volume tags using VolumeAttributesClass

## Expected behavior

VolumeAttributesClass working as intended

### LINKED ISSUE #649: EBS CSI Policy missing permissions for VolumeAttributesClass tagging
## Description

I would like to reopen closed [issue](https://github.com/terraform-aws-modules/terraform-aws-iam/issues/630) and prepare PR with removal or adjusting of [aforementioned condition](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts/policies.tf#L222-L229), if you don't mind, which blocks VolumeAttributesClass working.

- [x] ✋ I have searched the open/closed issues and my issue is not listed.

## Versions

- Module version [Required]: https://github.com/terraform-aws-modules/terraform-aws-iam/releases/tag/v6.6.1

- Terraform version: 1.12.2
- Provider version(s): aws 6.47.0

* EKS cluster v1.32+ with ebs-csi
* Create VolumeAttributesClass to overcome immutability limits of storage class
* ebs-csi won't be able to change tags on EBS with error: 

```
type="Warning" reason="VolumeModifyFailed" message="rpc error: code = Internal desc = Could not modify volume tags "vol-xx": operation error EC2: CreateTags, https response error StatusCode: 403, RequestID: id, api error UnauthorizedOperation: You are not authorized to perform this operation. User: <user> is not authorized to perform: ec2:CreateTags on resource: <resource> because no identity-based policy allows the ec2:CreateTags action.
```

Because condition in IAM policy https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts/policies.tf#L222-L229

## Expected behavior

ebs-csi able to change tags on EBS with VolumeAttributesClass

## Actual behavior

Error:

```
type="Warning" reason="VolumeModifyFailed" message="rpc error: code = Internal desc = Could not modify volume tags "vol-xx": operation error EC2: CreateTags, https response error StatusCode: 403, RequestID: id, api error UnauthorizedOperation: You are not authorized to perform this operation. User: <user> is not authorized to perform: ec2:CreateTags on resource: <resource> because no identity-based policy allows the ec2:CreateTags action.
```

### Terminal Output Screenshot(s)

None

## Additional context

Previously closed issue https://github.com/terraform-aws-modules/terraform-aws-iam/issues/630

somaz94:
## Description

The EBS CSI policy grants `ec2:CreateTags` only with a `ec2:CreateAction` condition (`CreateVolume` / `CreateSnapshot`), so the driver can tag volumes at creation but cannot tag existing volumes. This blocks `VolumeAttributesClass` tag modifications, which fail with `UnauthorizedOperation: ... ec2:CreateTags`.

This adds an opt-in `ebs_csi_volume_tagging` variable (default `false`). When enabled, the `ec2:CreateAction` condition is dropped so the driver can tag existing volumes/snapshots. The permission is intentionally gated behind an opt-in flag because the upstream EBS CSI driver omits it from its example policy on purpose, so existing users see no change by default.

Note `ec2:DeleteTags` in the same policy is already unconditional, so this restores the symmetry for tag modification.

## Motivation and Context

Fixes #649 (and the earlier, stale-closed #630).

## Breaking Changes

None. Default `false` preserves the current policy exactly.

## How Has This Been Tested?

- [x] `terraform fmt -recursive -check` — passes
- [x] `terraform validate` on `modules/iam-role-for-service-accounts` and `wrappers/iam-role-for-service-accounts` — both pass
- [x] `terraform-docs` regenerated; the submodule README inputs table matches
- [x] Verified default (`false`) renders the existing `ec2:CreateAction` condition, and `true` removes it


somaz94:
Still relevant — CI is green.

modules/iam-role-for-service-accounts/variables.tf
```diff
@@ -202,6 +202,12 @@ variable "ebs_csi_kms_cmk_arns" {
   default     = []
 }
 
+variable "ebs_csi_volume_tagging" {
+  description = "Determines whether to allow the EBS CSI driver to tag existing volumes/snapshots by removing the `ec2:CreateAction` condition on `ec2:CreateTags`. Required for `VolumeAttributesClass` tag modifications; disabled by default as it broadens tagging permissions"
+  type        = bool
+  default     = false
```
bryantbiggs, line 208: this default would automatically remove permissions which makes it a breaking change - must be `true` for backwards compatibility and users will have to opt into disabling

bryantbiggs (changes requested):

modules/iam-role-for-service-accounts/variables.tf
somaz94: Thanks for reviewing, and for merging `master` in.

I want to make sure I'm not talking past you, so I rendered the actual policy document three ways against this branch (`attach_ebs_csi_policy = true`, `terraform plan` → `terraform show -json` → the `aws_iam_policy` body). The variable's polarity is the opposite of what the name suggests at a glance:

**1. `master` vs this PR at the default (`ebs_csi_volume_tagging = false`) — byte-identical, zero diff:**

```
$ diff <(jq -S . master-policy.json) <(jq -S . pr-default-false-policy.json)
$ echo $?
0
```

The `dynamic "condition"` block uses `for_each = var.ebs_csi_volume_tagging ? [] : [1]`, so at the default it still emits the same `ec2:CreateAction` condition `master` emits unconditionally:

```json
{
  "Action": "ec2:CreateTags",
  "Condition": { "StringEquals": { "ec2:CreateAction": ["CreateVolume", "CreateSnapshot"] } },
  "Effect": "Allow",
  "Resource": ["arn:aws:ec2:*:*:volume/*", "arn:aws:ec2:*:*:snapshot/*"]
}
```

**2. `master` vs this PR with `ebs_csi_volume_tagging = true` — the condition is what gets *removed*:**

```diff
-      "Condition": {
-        "StringEquals": {
-          "ec2:CreateAction": [
-            "CreateVolume",
-            "CreateSnapshot"
-          ]
-        }
-      },
```

So `false` is the existing, narrower behavior and `true` is the opt-in that *broadens* `ec2:CreateTags` to already-existing volumes/snapshots (which is what `VolumeAttributesClass` tag modification needs). Defaulting to `true` would relax the policy for every current user on upgrade rather than preserve it — I think that's the inverse of what you were guarding against.

That said, your underlying requirement — **the default must be the backwards-compatible value, and any change must be opt-in** — is exactly right, and it's satisfied today. If the concern is that a `false` default *reads* like it's taking something away, I'm happy to flip the variable's sense so the default is literally `true`:

```hcl
variable "ebs_csi_restrict_tagging_to_create" {
  description = "Determines whether `ec2:CreateTags` is restricted to resource-creation via the `ec2:CreateAction` condition. Set to `false` to allow the EBS CSI driver to tag existing volumes/snapshots, which `VolumeAttributesClass` tag modifications require"
  type        = bool
  default     = true
}
```

Same rendered policy at the default, same opt-in semantics, and the default is `true` for backwards compatibility. Just say which you'd prefer and I'll push it — happy to go either way.

somaz94:
@bryantbiggs following up on the open thread here — I don't want to re-argue it, just to get a decision from you so it can move.

The short version of my Jul 29 reply: the default (`ebs_csi_volume_tagging = false`) renders **byte-identical** to `master`, so the backwards-compatibility requirement you raised is already met — `true` is the opt-in that broadens `ec2:CreateTags`, not the other way round.

You may still find that a `false` default *reads* like it removes something. If so, I offered to **flip the variable's polarity so the default is literally `true`**, which gets the same behavior with the more intuitive spelling. I'm happy either way — I just need to know which you'd prefer before touching it.

Also flagging that this repo's stale bot marks at 30 days of inactivity and closes 10 days later; last activity here was Jul 29, so it's due to be marked shortly. CI is green.

modules/iam-role-for-service-accounts/variables.tf
somaz94: Renamed it to `ebs_csi_restrict_tagging_to_create` with `default = true`, so the default renders exactly the same policy as master and nothing changes for existing users. If you need the driver to tag existing volumes (e.g. for `VolumeAttributesClass`), you opt out by setting it to `false`. Merged master in too, so it should be ready for another look.

PR #653, opened 2026-07-22
feat: Add AWS Load Balancer Controller AGA policy for v3.4

nickkadams:
## Description

Add a new `attach_load_balancer_controller_aga_policy` variable and IAM policy document to support AWS Global Accelerator (AGA) with the AWS Load Balancer Controller v3.4.

## Motivation and Context

The AWS Load Balancer Controller v3.4 ships a separate AGA controller IAM policy for Global Accelerator support. Upstream policy: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v3.4/install/aga_controller_iam_policy.json

## Changes

- New variable `attach_load_balancer_controller_aga_policy` (default `false`) in `variables.tf`
- New `data "aws_iam_policy_document" "load_balancer_controller_aga"` block in `policies.tf` matching all 7 statements of the upstream policy, with tag-based conditions and partition-aware ARNs
- New `data "aws_service_principal" "globalaccelerator"` data source for partition-aware service principal resolution
- Wired into `main.tf` (`policy_description`, `policy_name`, `source_policy_documents`)

## Breaking Changes

No.

## Testing

```hcl
module "aws_load_balancer_controller_irsa" {
  source = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts"

  role_name = "aws-load-balancer-controller"

  attach_load_balancer_controller_policy     = true
  attach_load_balancer_controller_aga_policy = true

  oidc_providers = {
    main = {
      provider_arn               = module.eks.oidc_provider_arn
      namespace_service_accounts = ["kube-system:aws-load-balancer-controller"]
    }
  }
}
```

PR #654, opened 2026-07-30
fix: Split LBC AGA policy into its own managed policy

nickkadams:
## Description

Enabling both `attach_load_balancer_controller_policy` and `attach_load_balancer_controller_aga_policy` currently merges both policy documents into a single customer managed IAM policy (`aws_iam_policy.this`). This change attaches the AGA policy as its own standalone `aws_iam_policy`/`aws_iam_role_policy_attachment` pair instead, so it no longer shares a size budget with the base Load Balancer Controller policy. Updated the `load_balancer_controller_aga_irsa` example to enable both flags together, reproducing the scenario.

Since the AGA policy is now its own resource rather than folded into the combined policy, it also gets its own overridable `load_balancer_controller_aga_policy_name` and `load_balancer_controller_aga_policy_description` variables, defaulting to `AWS_LBC_Global_Accelerator` and the existing description directly on the variable defaults, and passed through the wrapper module.

## Motivation and Context

The AGA policy was introduced in #653. When both flags are enabled together, the merged policy document can exceed the 6,144 character quota for a customer managed IAM policy, causing:

```
Error: Cannot exceed quota for PolicySize: 6144
```

## Breaking Changes

No. `attach_load_balancer_controller_aga_policy` behaves the same from the caller's perspective; it now provisions a second `aws_iam_policy`/`aws_iam_role_policy_attachment` pair under the hood instead of folding into the combined policy. The new `load_balancer_controller_aga_policy_name`/`load_balancer_controller_aga_policy_description` variables are optional and additive.

## How Has This Been Tested?


nickkadams:
Hi @bryantbiggs, just following up here. Would you mind confirming this approach looks right when you get a chance, and helping get it merged?

Separately, I noticed `terraform-aws-eks-pod-identity` has a similar AGA policy PR (terraform-aws-modules/terraform-aws-eks-pod-identity#55) that stalled and got auto-closed. The approach there already matches this module's pattern (standalone policy per feature), so I'd like to revive it with a minor partition fix (`arn:${local.partition}:globalaccelerator::*:...` instead of the hardcoded `arn:aws:...`, for GovCloud/China support) once this one lands.

modules/iam-role-for-service-accounts/main.tf
```diff
@@ -34,14 +33,23 @@ locals {
...
-    var.attach_load_balancer_controller_aga_policy ? "AWS_LBC_Global_Accelerator" : null,
     var.attach_amazon_managed_service_prometheus_policy ? "Amazon_Managed_Service_Prometheus" : null,
     var.attach_node_termination_handler_policy ? "Node_Termination_Handler" : null,
     var.attach_velero_policy ? "Velero" : null,
     var.attach_vpc_cni_policy ? "VPC_CNI_${var.vpc_cni_enable_ipv4 ? "IPv4" : "IPv6"}" : null,
     var.name,
     "default"
   ))
+
+  load_balancer_controller_aga_policy_name = coalesce(
```
bryantbiggs: make this the default for the policy name variable

modules/iam-role-for-service-accounts/main.tf
```diff
@@ -34,14 +33,23 @@ locals {
...
+    var.load_balancer_controller_aga_policy_name,
+    "AWS_LBC_Global_Accelerator"
+  )
+
+  load_balancer_controller_aga_policy_description = coalesce(
```
bryantbiggs: make this the default for the policy description

bryantbiggs:
whoops, sorry I am a bit behind on reviewing these days. two small changes and we can probably get it merged. it will cause a blip in permissions but since its applied to a controller, it should be fine and go through as a no-op on functionality 

feel free to apply the same changes/PR to the EKS Pod Identity repo and I can get that in as well

modules/iam-role-for-service-accounts/main.tf
nickkadams: Done — defaulted directly on `load_balancer_controller_aga_policy_name` in variables.tf.

modules/iam-role-for-service-accounts/main.tf
nickkadams: Done — defaulted directly on `load_balancer_controller_aga_policy_description` in variables.tf.

bryantbiggs (approved):
thank you!

antonbabenko:
This PR is included in version 6.8.1 :tada:

PR #657, opened 2026-08-06
fix: Add output guards

makarov-roman:
Hi guys,
I was testing apply/destroy for eks project which uses the module and noticed that destroy order for EBS is broken.

I use arn output to pass the role into ebs addon. And it means that terraform only sees arn as ref. When you destroy (or replace), you'll have the situation when the role still exist and waits for workload to be gracefully deleted, but attachment and policy are gone.

```
Case A — current module (no depends_on):
  destroyed-attachment -- permissions gone
  destroyed-policy
  destroyed-workload   <-- workload still running, failing to perform finalizers
  destroyed-role

  Case B — with depends_on on the output:
  destroyed-workload   <-- drains first
  destroyed-attachment
  destroyed-policy
  destroyed-role
```

the same problem is applicable to pod-identity module (I haven't migrated yet)

Thanks!

### DIFF
--- a/modules/iam-role-for-service-accounts/outputs.tf
+++ b/modules/iam-role-for-service-accounts/outputs.tf
@@ -5,11 +5,23 @@
 output "arn" {
   description = "ARN of IAM role"
   value       = try(aws_iam_role.this[0].arn, null)
+
+  depends_on = [
+    aws_iam_role_policy_attachment.this,
+    aws_iam_role_policy_attachment.additional,
+    aws_iam_role_policy.inline,
+  ]
 }
 
 output "name" {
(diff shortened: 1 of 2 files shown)


bryantbiggs:
thank you but unfortunately we are not going to accept this change at this time

makarov-roman:
@bryantbiggs sorry, but can you elaborate? 
such approach is already used in the code base - [e.g. here](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/d386adc021dc370efe1bceb7991fbed6e9787c84/outputs.tf#L9)

or you don't think the problem is worth solving? or you don't like the implementation itself?

